PR #5, opened 2018-07-27
Upgrades to DefConstructor, properties

dginev:
(while fleshing out sectioning infrastructure)

This is a work in progress and may not even compile on individual commits, as there are too many moving pieces.

Key goals (for my own memory):
 * Allow dual-typed signatures for `DefConstructor!` replacements, where strings are compiled and closures are accepted as-is.
 * Similarly allow a dual-typed HashMap vs closure for properties.

After this the PR can be merged and the sectioning support flesh-out can continue on a new branch.

### DIFF
--- a/rtx/src/lib.rs
+++ b/rtx/src/lib.rs
@@ -6,6 +6,7 @@
 #![allow(unused_variables, unused_mut, unused_macros)]
 #![cfg_attr(feature = "clippy", feature(plugin))]
 #![cfg_attr(feature = "clippy", plugin(clippy))]
+#![feature(macro_literal_matcher)]
 
 #[macro_use]
 extern crate lazy_static;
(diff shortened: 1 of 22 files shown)


dginev:
Another useful upgrade to mention (already in master) is the `unpack!` and `unpack_to_string!` macros which are early attempts at emulating the simple:

```perl
my ($a,$b,$c) = @_;
```

dynamic behavior from Perl. The key is to provide an identical syntactic ease for binding authors, while keeping the Rust strictness underneath. So the current PR introduces:

```rust
unpack!(args => a,b,c);
```

Although it could equally well be ```assign!(a,b,c = args)``` etc.

dginev:
PR goal is achieved at this point, merging.

PR #9, opened 2018-08-26
performance upgrade: Rc-wrapped atomic digested objects

dginev:
Base idea: 
 * Avoid deep clones on `Digested` objects, but resign to needing clones for at least the atoms, as they cohabitate all of 1) state storage, 2) main application flow, 3) `node_boxes` map that associates constructed nodes to their digested boxes.
 * Also, use as lightweight wrappers as possible to improve performance,
 * So Digested has now evolved into a Rc-pointer for Tbox and Whatsit, where Whatsit also requires a RefCell as we still want to allow mutability.

```rust
pub enum Digested {
  TBox(Rc<Tbox>),
  Whatsit(Rc<RefCell<Whatsit>>),
  List(List),
  Postponed(Rc<Tokens>),
}
```

Also, using Rc internally to the digested variants that end up cloned frequently allows to more freely store digested objects as-is (similarly to the rust-libxml Node objects), as the underlying data remains shared between multiple Digested enum instances.

### DIFF
--- a/rtx/src/converter.rs
+++ b/rtx/src/converter.rs
@@ -165,7 +165,7 @@ impl Converter {
       Err(e) => {
         // TODO digestion failed, report
         e.log_fatal();
-        Digested::List(Box::new(List::new(Vec::new())))
+        Digested::List(List::new(Vec::new()))
       },
       Ok(d) => d,
     };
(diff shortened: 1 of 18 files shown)


dginev:
The one real question is wether lists should also be Rc-wrapped. reason being that lists can nest within each other, so a top-level clone may end up too expensive.

The current redesign already offers another 10-15% speedup however.

PR #10, opened 2018-08-28
Cow<str> contents of Token when static references

dginev:
Token construction can be further optimized by avoiding owning into the heap `&'static str` declarations.

It's a bit tricky to evolve the macro language to support this, as it requires a way to distinguish between static and non-static lifetimes. A "workaround" is to use the new unstable `:literal` macro designator, as to allow storing literal strings as borrowed, and owning any others.

However, there is additional headache involved with (seemingly?) nesting macros when the innermost one has a `:literal` definition, as the parent rtx module refuses the compile, even with the feature annotation... Maybe something to revisit when the designators are stabilized in future Rust versions.


dginev:
Oh, this PR only builds half-way on travis, since the `:literal` feature is nightly-only.

dginev:
There seems to have been a notable performance improvement from master to this PR, here is a comparison:

master [before]
master_perf

Token-Cow-refactor [after]
main_cow_token

dginev:
Looks like a 10% improvement in overall instructions, and also an improvement in Data numbers by about as much. Tempting to merge.

dginev:
For reference, here is the kcachegrind screenshot of the state at 0.1.3, before any of the optimization work

before_optimization

PR #11, opened 2018-09-03
Rc<String> via Crow for immutable Token text

dginev:
So far filing this under the `failed experiments` heading, but adding for reference (and maybe future brainstorms).

I "upgraded" the Token text field to hold a new datastructure `Crow` which is a (supposedly) lightweight version of Cow, but instead of ownership it has shared reference via an Rc. The true "holy grail" here would have been to redesign Token into a lifetime-carrying struct, where text is just a `&str`. However, that involves a singificant redesign and seems almost intractable (and exists on a separate branch as I experiment).

In this PR the workaround is easier to write into compiling code, but:
 * For best ergonomics we need a global HashMap in the `crow` module that holds the String memory arena.
 * For this to work with `lazy_static` we need each value to be the slower `Arc` pointer (thread-safe) and have the hashmap encapsulated behind a `Mutex`.

I am yet to understand the full performance benchmark implications, but it looks like there was a ~5-10% slow down in effect, which is not entirely surprising given all the new overhead I end up introducing. And this is on a benchmark test that is excessively redundant and one could expect a very large number of repeated `Token.text` values ... Will probably close without merging the experiment and give up on this optimization front for now.


dginev:
Put differently,  this is a trade-off question as to whether cloning a string (with all implied memory allocations) is faster and preferable to looking up (and reference counting) a single memory allocation, via the mutex-hashmap-arc route.

It is quite clear that having bare immutable references into `&str` objects will be superior in both performance and memory to these acrobatics though... Hm.

PR #27, opened 2019-01-23
Refactor to using proc-macro2 and newest quote crate

dginev:
This PR allows for simplifying some of the code generation infrastructure, although it wasn't an easy climb to get to implement the `ToTokens` trait in the new proc-macro2 abstractions (mostly due to sparse documentation/examples).

I'm hoping staying current with the ecosystem will give us the best chance to implement the auto-binding of state for in-closure bindings.

But also nicely - a lot of cargo cult code can now be dropped as the official Rust macro infrastructure has matured.


dginev:
I had a major conceptual breakthrough today -- though in retrospect it seems like the silliest breakthrough one can have.

Turns out I can use the fact the "custom derive" crates are turing-complete and just "wing it" for the context tracking information. Rust really wants to disallow any technique that does contextual variable capture, and for good reason, but I think the design I landed at is perfectly safe and sound for our application (latexml bindings).

I added a couple of macros that expanded into "nothing" (the empty TokenStream) and were purely side-effect. The side-effects in question -- mutate a global static variable (unsafely!) inside the custom derive crate, to increment and decrement a "context depth".

Once that was present and introduced to all binding sub-closures (think `before_digest` and friends), it became transparent to the Custom Derive crate whether it was called from the main binding context (and should use the `outer_state!()` value), or was called inside the closure context (and should use the `inner_state!()` value, bound by the current closure macros). Note that this is all happening at the macro expansion stage in compilation, so it's entirely invisible for the main compile pass and the runtime.

I extracted a gist that summarizes the technique here:
https://gist.github.com/dginev/63447ca98d0974f5c67f5ac646965dd4

Ready to merge and start simplifying the currently implemented macros, at which point I believe another minor release will be in order.

PR #28, opened 2019-01-24
Allow for perl-brevity in binding definitions

dginev:
Only augmented for DefMacro at the start of this PR, but even so! This is now valid Rust code in rtx:

```rust
DefMacro!("\\a", sub{ DefMacro!("\\b", "c"); });
```

I would like to upgrade the rest of the bindings that support the `sub[]` notations, and ensure that there is a proper `.into_*_result` trait for each of them, at which point this brevity can become the expected norm in rtx.

### DIFF
--- a/rtx_core/src/common/store.rs
+++ b/rtx_core/src/common/store.rs
@@ -486,28 +486,3 @@ impl<'a> From<&'a Stored> for Token {
     }
   }
 }
-
-pub trait IntoOption<T>: Sized {
-  /// Performs the conversion.
-  fn into_option(self) -> T;
-}
-
-impl<'a> IntoOption<Option<String>> for &'a str {
-  fn into_option(self) -> Option<String> { Some(self.to_string()) }
-}
-
-impl<T> IntoOption<Option<T>> for Option<T> {
-  fn into_option(self) -> Option<T> { self }
-}
-
-impl IntoOption<bool> for bool {
-  fn into_option(self) -> bool { self }
-}
-
-impl<T> IntoOption<Option<Vec<T>>> for Vec<T> {
-  fn into_option(self) -> Option<Vec<T>> { Some(self) }
-}
-
-impl<T> IntoOption<Option<VecDeque<T>>> for VecDeque<T> {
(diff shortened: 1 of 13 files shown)


dginev:
This is now a real Rust binding, and I'm very excited!

```rust
DefPrimitive!("\\obeyspaces", sub {
   AssignCatcode!(' ', Catcode::ACTIVE);
   LetI!(&T_ACTIVE!(" "), T_CS!("\\space"));
});
```

compare it to the original Perl:
```perl
DefPrimitiveI('\obeyspaces', undef, sub {
    AssignCatcode(" " => 13);
    Let(T_ACTIVE(" "), '\space');
    return });
```

It is indeed a little annoying that there are so many exclamation marks, but it helps to remember what they alias. Here is a capture of the result after the macro expansion stage, as provided by the `cargo expand` utility. Which then gets turned into very optimized bytecode, courtesy of LLVM optimizations.

```rust
{
    let st: &mut State = {
        struct _Bound;
        state
    };
    {
        let (cs, paramlist) = parse_prototype("\\obeyspaces", st)?;
        {
            def_primitive(
                cs,
                paramlist,
                Rc::new(|stomach, whatsit, inner_state| {
                    {
                        {
                            struct _SFrame;
                        };
                        let macro_out = {
                            {
                                let st: &mut State = {
                                    struct _Bound;
                                    inner_state
                                };
                                st.assign_catcode(' ', Catcode::ACTIVE, None)
                            };
                            {
                                let st: &mut State = {
                                    struct _Bound;
                                    inner_state
                                };
                                st.let_i(
                                    &Token {
                                        text: Cow::Borrowed(" "),
                                        code: Catcode::ACTIVE,
                                    },
                                    Token {
                                        text: Cow::Borrowed("\\space"),
                                        code: Catcode::CS,
                                    },
                                    None,
                                )
                            };
                        };
                        {
                            struct _EFrame;
                        };
                        macro_out
                    }
                    .into_digested_result()
                }),
                PrimitiveOptions::default(),
                st,
            );
        };
    }
};
```

PR #29, opened 2019-01-25
TeX.pool coverage 3

dginev:
So far here:
 * added `\meaning` macro and flesh out related stringifying infrastructure and type interfaces. Pretty good macro for stress-testing parameters.
 * exhaustively covering TeX.pool line-by-line, reached end of `\meaning` macro, so nearing line 900 in the original file

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ sudo: false
 dist: trusty
 language: rust
 rust:
+  - stable 
   - nightly
 addons:
   apt:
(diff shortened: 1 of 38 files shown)


dginev:
Ok, this PR Is now completed for merge, I believe it contains enough:
 * First `1101` lines of TeX.pool.ltxml are now reimplemented in Rust
 * Took some (sideways?) steps towards the last bit of binding ergonomics, now using redundant idioms such as `before_digest => before_digest!(..)`
 * simplified some of the earliest code that predated Rust 1.30 and non-lexical lifetimes, to fully leverage NLL -- Rust now allows a lot more lenience with mutability, as it can understand more of the safe patterns (of the kind "but I am not using this ever again after, why are you complaining?!")
 * Some steps to properly using the state `stash` and scoping mechanisms.

PR #32, opened 2019-02-20
PASS digestion/defaultunits

dginev:
Pass the defaultunits test.

 * Fleshes out more of the Glue and Register infrastructure
 * `\newcommand` and `TeXFileName` improved implementations
 * Added a big chunk of `RawTeX` definitions for `latex_other_in_appendices.rs`

One observation is that introducing significant snippets of RawTeX definitions visibly slows down the test run. Not terribly, a change between `0.48s` to `0.51s` for the entire test suite, but it can be noticed in the timed runs.


rtx_package/src/package/pool/latex_other_in_appendices.rs
```diff
@@ -269,174 +269,174 @@ LoadDefinitions!(state, {
...
+    \newif\ifmath@fonts \math@fontstrue
+    \newbox\@labels
+    \newif\if@inlabel \@inlabelfalse
+    \newif\if@newlist   \@newlistfalse
+    \newif\if@noparitem \@noparitemfalse
+    \newif\if@noparlist \@noparlistfalse
+    \newif\if@noitemarg \@noitemargfalse
+    \newif\if@nmbrlist  \@nmbrlistfalse
+
+    \def\glb@settings{}%
```
dginev, line 437: This whole snippet seems to work just fine now! rtx++

PR #36, opened 2019-03-02
First Setup: encoding tests

dginev:
Draft PR to keep an eye on the file diffs as I get closer to the fully operational encoding dance.

Tasks:
- [x] Refactor binding dialect to always require a `&mut Stomach` argument in loading definitions, thus enabling `ProcessOptions!(stomach, state)`
- [x] Refactor binding dialect to a convention where `state` is the last argument (whenever possible in macro signatures)
- [x] Precise parity for `Document::insert_pi` attribute order
- [x] Encoding refactor for file Mouths, setup buffered reads with decoding step
- [x] Refactor binding-related package submodules into the `rtx_package::package::api` sub-namespace, to make it a little more manageable
- [ ] dev aid: would be nice to fully implement `Locator` and integrate source file information in the errors, especially since loading the encoding `.def` files inititally generates a lot of unknown macro errors (as they should, a lot of the pool files are yet to be ported over).
- [ ] Add `fontenc.sty.ltxml`, `inputenc.sty.ltxml` and `textcomp.sty.ltxml` implementations
   - [x] binding skeleton and basic macros
   - [x] Flesh out full `InputDefinition` infrastructure
   - [x] including `kpathsea`-based file search
   - [ ] Port latex package option processing
   - [ ] Port alignments `&`
   - [ ] Port `{tabular}`
- [x] minor: fix `\box` and `\copy` bindings to work at the primitive stage


rtx_package/src/package/api/content.rs
```diff
@@ -0,0 +1,687 @@
...
+      }
+    },
+    None => false,
+  };
+  state.extra_bindings_dispatch = taken_dispatcher;
+
+  Ok(is_contrib)
+}
+
+/// TODO: Flesh out with the full infrastructure, incremental functionality for now.
```
dginev, line 61: I now have space to do a deep dive and port over the full `InputDefinition` logic, while remaining just under 1000 lines in this new `rtx_package::package::api::content` module.

dginev: Though I'm happier about setting up the first submodules, since I could now split out into ever smaller dependencies, as confusion levels rise again

dginev:
I've published a [kpathsea wrapper](https://crates.io/crates/kpathsea) today, and can already use it in rtx via bb3d9a6 , which prints:

```
Warn:rtx_package::package::api::content Hey! Found with kpsewhich: "/usr/share/texlive/texmf-dist/tex/latex/base/ansinew.def"
```

which is lovely. Oh, and since we're doing C-level linking, it's FAST and free of any system calls!

dginev:
Renaming here with the idea for a merge into master for the bits already fleshed out, as they keep the rest of the test suite in tact, and could help minimize the diff for reviewing the core encoding upgrades.

dginev:
Continues in #37

PR #37, opened 2019-03-05
Locator, Error, Object infrastructure; low-level encoding bindings

dginev:
Draft PR to keep an eye on the file diffs as I get closer to the fully operational encoding dance. Continuing from #36

Tasks:
- [x] Refactor binding dialect to always require a `&mut Stomach` argument in loading definitions, thus enabling `ProcessOptions!(stomach, state)`
- [x] Refactor binding dialect to a convention where `state` is the last argument (whenever possible in macro signatures)
- [x] Precise parity for `Document::insert_pi` attribute order
- [x] Encoding refactor for file Mouths, setup buffered reads with decoding step
- [x] Refactor binding-related package submodules into the `rtx_package::package::api` sub-namespace, to make it a little more manageable
- [x] dev aid: would be nice to fully implement `Locator` 
- [x] dev aid: and integrate source file information in the errors, 
   * especially since loading the encoding `.def` files inititally generates a lot of unknown macro errors  
   * (as they should, a lot of the pool files are yet to be ported over).
- [ ] Add `fontenc.sty.ltxml`, `inputenc.sty.ltxml` and `textcomp.sty.ltxml` implementations
   - [x] binding skeleton and basic macros
   - [x] Flesh out full `InputDefinition` infrastructure
   - [x] including `kpathsea`-based file search
   - [x] Port latex package option processing
   - [ ] Port alignments `&`
   - [ ] Port `{tabular}`
- [x] minor: fix `\box` and `\copy` bindings to work at the primitive stage

### DIFF
--- a/rtx/src/main.rs
+++ b/rtx/src/main.rs
@@ -1,4 +1,5 @@
-use log::*;
+#[macro_use]
+extern crate rtx_core;
 use rtx_core::common::{Config, DataSize, OutputFormat};
 use rtx_package::converter::Converter;
 use std::env;
@@ -7,7 +8,8 @@ use std::rc::Rc;
 
 fn main() {
   if rtx_core::util::logger::init(log::LevelFilter::Info).is_err() {
-    error!("Failed to load logger, aborting early. Please check rtx_core::util::logger installed correctly.")
+    Error!("rtx", "logger", None, None, 
+      "Failed to load logger, aborting early. Please check rtx_core::util::logger installed correctly.");
   }
   let mut argv = env::args();
   argv.next();
@@ -15,7 +17,7 @@ fn main() {
   let source = match argv.next() {
     Some(s) => s,
     None => {
-      error!("Please provide a source document! Exiting...");
+      Error!("rtx","", None, None, "Please provide a source document! Exiting...");
       process::exit(1);
     },
   };
@@ -32,7 +34,8 @@ fn main() {
   };
   let mut converter = Converter::from_config(opts.clone());
   if let Err(e) = converter.prepare_session(&opts) {
-    error!("Could not prepare converter session! : {}", e);
+    let message = s!("Could not prepare converter session! : {}", e);
+    Error!("rtx","session",None, None, message);
     process::exit(1);
   }
   // Perform the conversion:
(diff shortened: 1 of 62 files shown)


dginev:
A bit trickier than I would wish to invent the improved error-reporting, but definitely worth approaching the latexml levels of metadata. The last commit now uses the locators, so tying together the whole body of the PR with this screenshot:

error_undefined_after

which can now be produced via the code:
```rust
use rtx_core::common::error::*;
// ...
Error!("unexpected", category, self, &message, state);
```

And here is the same error in latexml for comparison of things left to add:
```
Error:undefined:\DeclareFontEncoding The token T_CS[\DeclareFontEncoding] is not defined.
	at t1enc.def; line 48 col 20 - line 48 col 20
	Defining it now as <ltx:ERROR/>
	In Core::Stomach[@0x55834f7265e0] /usr/local/texlive/2017/texmf-dist/tex/latex/base/t1enc.def; from line 48 col 20 to line 48 col 20
	 <= Core::Gullet[@0x55834f72cfd8] <= Core::Definition::Constructor[\usepac... <= Core::Stomach[@0x55834f7265e0] <= ...
```

dginev:
With an hour of extra tweaking I have arrived at:
```rust
      let message = s!("The token {} is not defined.", token.stringify());
      Error!("undefined", token, self, state, &message, "Defining it now as <ltx:ERROR/>");
```

producing:
logged_error

I would say this already feels "good enough" for this stage of work, as it gives me the full location fingerprint of where in the source, and where in the code, the error took place. It's also very close to latexml in the message printed, as well as in the macro ergonomics.

So I'll refactor my use of the logging facade towards the new capitalized `Info!, Warn!, Error!, Debug!` sublanguage, and move on.

dginev:
The basic loading and TeX interpretation of the encoding `.def` files seems to be mostly operational now.

The next (and hopefully near-last) hurdle in interpreting the encoding tests is actually getting the `{tabular}` framing for them working. As it turns out, that requires the full alignment infrastructure, so strangely enough the `30_encoding` test set is strongly dependent on the `53_alignment` test set passing beforehand.

I will merge the PR as tihs bit of infrastructure is valuable and ready, and reconsider if the next push should be towards a full alignment focus. Still waiting on the `\halign` refactor to drop in latexml, before I start on that bit of the code base.

PR #40, opened 2019-03-11
PASS structure/abstract

dginev:
Passing the abstract.tex test:
 * add missing frontmatter and article.cls pieces
 * Figured out that `stringify` is a good-enough identifier, and I can directly mimic latexml's Definition::stringify in the rust Definition trait, which made life better
 * Turned `Parameters` into a thin wrapper over a `Vec`, akin to `Tokens`, to be more idiomatic
 * Ported over `AddToMacro!`
 * KeyVals, KeyVals, Keyvals

The big remaining part of this PR is to flesh out the needed parts of the very impressive KeyVals interface.


dginev:
Ok, I managed to add the first stab at a (full) scaffolding for KeyVals, with the basic functionality working and a lot of open places to jump into towards the full support, as we port over more sophisticated bits of latexml.

I also managed to quite neatly recreate the `@LaTeXML::LIST` technique by using the stomach object and `.regurgitate()`, all in all quite happy with the progress here.

Added some finer fixes and `{abstract}` test passes successfully, with correctly passed in and serialized metadata.

Two big performance wins that happened largely "by the way" of working here that are worth noting: 
  * `Tokenize!`, `TokenizeInternal!`, `RawTeX!` -- containing only "string literal arguments" -- are now tokenized to rust objects during compile time. Hence the runtime is now reduced to a `stomach.digest` call on the pre-tokenized objects, which shaves off a neat amount of overhead. 
   * The `Store` enumeration now stores parameters behind an `Rc<>` smart pointer, which visibly sped things along, and was a necessity for making keyval definitions work efficiently.

Will review quickly and merge here!

PR #46, opened 2019-03-18
Fluent binding ergonomics

### LINKED ISSUE #24: Fluent binding ergonomics
The just merged #23 has fully set the stage to attack the `Def*` mini language with a goal of full parity with the latexml `.ltxml` definitions. 

It is now possible to hide `state` entirely, AND to invoke nested binding constructs, passing a stomach where needed.

The steps remaining that I can see:
 * [x] auto-bind the closest state in closures such as `before_digest`
 * [x] refactor existing bindings to never use the explicit state form
 * ~~remove the explicit state form entirely, in favor of using local scope redefinitions of the `state!()` macro, which are less prone to human error, and will simplify `setup.rs` significantly.~~
 * [x] compile-time upgrade: move the heavy load definitions into dedicated rust functions, such as the existing `fn def_macro`. 
 * [x] Biggest remaining pain-point in writing down definitions: there is a need for better argument parsers for dependent closures. In particular, it would be great if `before_digest => before_digest!(...)` could become simply `before_digest!(...)`

dginev:
Closes #24 .

This PR offers a major refactoring of `DefConstructor`, `DefPrimitive`, `DefEnvironment` and `DefMacro`, covering the biggest payload of bindings already ported to rtx.

I have removed the last of the visible rust-y boilerplate from the macros, and can now showcase an honest equivalent of the most complex of latexml binding definitions.

Take three examples from Perl, and their current forms in Rust:
```perl
DefPrimitiveI('\nullfont', undef, undef, font => { family => 'nullfont' });

DefEnvironment('{itemize}',
  "<ltx:itemize xml:id='#id'>#body</ltx:itemize>",
  properties => sub { beginItemize('itemize', '@item'); },
  beforeDigestEnd => sub { Digest('\par'); },
  locked => 1, mode => 'text');


DefConstructor('\hbox BoxSpecification HBoxContents', sub {
    # ... snip ...
  },
  mode => 'text', bounded => 1,
  sizer => '#2',
  beforeDigest => sub { reenterTextMode(); },
  afterDigest => sub {
   # ... snip ...
 });
```

into:
```rust
DefPrimitive!("\\nullfont", None, font => {family => "nullfont"});

DefEnvironment!("{itemize}",
  "<ltx:itemize xml:id='#id'>#body</ltx:itemize>",
  properties => { BeginItemize!("itemize", "@item") },
  before_digest_end => { Digest!("\\par")?; },
  locked => true,
  mode => "text");


DefConstructor!("\\hbox BoxSpecification HBoxContents", sub[document, args, props, state] {
    // ...snip...
  },
  mode => "text",
  bounded => true,
  sizer => "#2",
  before_digest => sub[stomach, state] {reenter_text_mode(false, state)},
  after_digest => sub[stomach, whatsit, state] {
    //...snip...
});
```

### Features
So, what are the *fluent* aspects that we now have, and I consider here to stay:
 * Fully typed interface -- a `true` is a boolean, while `1` is a number. The Rust compiler will help whenever e.g. a `stomach` is used as a `gullet` in a hurry (or when moving code around between phases, which happens a lot in binding development of e.g. sTeX-complexity systems).
 * Brief and Perl-near syntax, with the capacity to get briefer / more flexible. I already made it flexible to use rust-style separators (`a: b`) and perl-style separators (`a => b`) for the key-value pairs in the bindings, a lot more is possible.
 * Localized and predictable objects - state, stomach, etc are local to each closure call, no magical mutable globals exist, which makes it very reliable to reason about the code snippets in isolation from the rest of latexml.
 * Compile-time optimizations and translation into safe Rust code

The big goal is to *copy-paste* from latexml bindings, adapt as little as possible, and arrive at faster+safer code. I think this is now very real.

### Implementation
And on a technical note, the 4 key features that allowed for all of this macro machinery to come to life:
 * Implicitly binding a local `state` via the compiler extension `rtx_codegen` sub-crate, and the `bind_state!()` family of macros, allows to write brief+smart auxiliary macro calls (e.g. `Let!("\a","\b");`).
 * The big refactor of the current PR allowed for the fully flexible options syntax one expects from a Perl hash -- while still translating into fully typed Rust closures behind the scenes, courtesy of what I've learned has been named the `token-tree munchers` macro pattern [docs](https://danielkeep.github.io/tlborm/book/pat-incremental-tt-munchers.html).
* auto-wrapping into Store-able type unions makes it possible to write `font => {family => "nullfont"}` and have Rust auto-cast the various components into the types expected for internal processing (notably `Store` and `Option` wrappers).
* And last but definitely not least, smart use of the "macro designators" allowed to write smarter binding rules. Being able to distinguish between `literal`, `ident`, `block` and `expr` allows for much more flexibility in the binding machinery. E.g. we can compile `literal` replacements and expansions into closures/tokens objects and eliminate overhead, we can auto-wrap function signatures around `block`s, etc.

The rest is elbow grease and trial-and-error. Really relieved to see the binding language as "solved", it was the first big challenge I tackled in rtx, and it took some maturity of Rust itself before I could pull it all off.


dginev:
Also to note, a bunch of constructs are yet to be refactored -- `Tag`, `DefRegister`, and the internal variants of all constructs such as `DefConstructorI`. Will probably refactor those before merging here, will see if I can enlist some feedback beforehand also...

brucemiller:
Think of the non-internal Def forms as simply macros that expand into the internal forms doing in general only 2 things: (1) checking for valid options and (2) splitting and parsing the "prototype" into the CS being defined, and it's parameter list.  You've probably already got the code for all those things, so covering the internal forms should just be a minor bit of code shuffling (he said glibly).

dginev:
Thanks for the comment! And yes, happy to say most of the code is there, just needs a big sweep after the final upgrades.

For (1) that is a bit of logic that I have retooled in Rust, because the compiler does it for us for free! The idea being that you create option structs such as:
```rust 
pub struct ExpandableOptions {
  pub locked: bool,
  pub protected: bool,
  pub outer: bool,
  pub long: bool,
  pub scope: Option<Scope>,
  pub alias: Option<String>,
}
```
and then have the `DefMacro!` call build an instance of that struct internally and leave it to rust to monitor we have the right types of values, as well as only allowed values.

(2) is really the part that make internal forms of the macros necessary in Rust, and that only because I can't find a reliable way to distinguish between the various argument forms... In fact, it may be possible to have a single `DefMacro!` that accepts the unparsed prototype form, and the parsed `cs,parameters` argument form, entirely by monitoring the types of rust objects passed in... 

To take the briefest example from above, the main form may look like:
```rust
DefPrimitive!("\\nullfont", None, font => {family => "nullfont"});
```
while the internal form may be just an extra `None`:
```rust
DefPrimitive!("\\nullfont", None, None, font => {family => "nullfont"});
```
standing in for "there are no parameters here, don't parse them out."
I could fish around for a path merging them, since reducing the extra macro boilerplate is quite good for sanity... Much easier to maintain a 500 line macro setup than a 2000 line one!

dginev:
Ok, this really shows I got the approach right! I have successfully made obsolete all internal argument forms, while adding their functionality to the main macro versions. 

So now there is even more "do what I mean" in Rust than in Perl! Ha! The macro machinery will understand whether the use of the macro implies parsing a prototype, or already provides the `$cs` and `$paramlist` explicitly, and will take the corresponding path.

Pretty cool! And now this PR is fully ready to merge. `rtx` has fluent binding ergonomics!

PR #70, opened 2021-06-25
Add 'native_kpsewhich' feature

tkw1536:
Previously this project required the 'libkpathsea' library in order to compile properly. This is not an issue on sufficiently new versions of Ubuntu, but was a problem on OS X machines.

Instead of using kpsewhich form the kpathsea library, it is possible touse the 'kpsewhich' executable at runtime. 
This PR implements this approach and hides the previous approach behind a 'native_kpsewhich' feature flag.

### DIFF
--- a/rtx_core/src/util/pathname.rs
+++ b/rtx_core/src/util/pathname.rs
@@ -297,7 +297,7 @@ pub fn extension(pathname: &str) -> String {
 
 pub use kpsewhichimpl::kpsewhich as kpsewhich;
 
-#[cfg(feature = "native_kpsewhich")]
+#[cfg(feature = "kpsewhich_libkpathsea")]
 mod kpsewhichimpl {
   use kpathsea::Kpaths;
   pub fn kpsewhich(candidates: &[&str]) -> Option<String> {
@@ -311,7 +311,7 @@ mod kpsewhichimpl {
   }
 }
 
-#[cfg(not(feature = "native_kpsewhich"))]
+#[cfg(not(feature = "kpsewhich_libkpathsea"))]
 mod kpsewhichimpl {
   use std::process::Command;
   use std::string::String;
--- a/rtx_core/src/util/pathname.rs
+++ b/rtx_core/src/util/pathname.rs
@@ -1,4 +1,3 @@
-use kpathsea::Kpaths;
 use lazy_static::lazy_static;
 use regex::Regex;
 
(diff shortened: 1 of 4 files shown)


dginev:
Since I maintain the kpathsea wrapper, I'd rather figure out how to make the library available on MacOS, rather than add system call overhead... PR is pretty good though!

rtx_core/src/util/pathname.rs
```diff
@@ -295,16 +294,63 @@ pub fn extension(pathname: &str) -> String {
...
+    for candidate in candidates {
+      if let Some(path) = exec(candidate) {
+        return Some(path);
+      }
+    }
+    None
+  }
+
+  // executable name to be used for kpsewhich
+  static KPSEWHICH_EXECTUABLE: &str = "kpsewhich";
```
dginev, line 329: You ought to check if kpsewhich is available in runtime, using [which](https://docs.rs/which/4.1.0/which/), rather than assume it will be permanently available as a compile-time flag. texlive can appear and disappear while rtx is installed, so one can't rely on it too much...

... which is why I was so happy the kpathsea library existed separately from texlive, so that one could rely on it being there all the time.

brucemiller:
> ... which is why I was so happy the kpathsea library existed separately 
> from texlive, so that one could rely on it being there all the time.

Only in a statically linked binary.

It's nice to imagine a fall-back to the executable if the library is not 
available (presumably at compile time; handling a runtime linker error 
doesn't seem feasible).  It does raise a future question about 
distributing "binaries", rather than requiring compile from source.

dginev:
> Only in a statically linked binary.

Well yes and no. I did not mean to imply that I can guarantee that kpathsea will be always present as a distributor *of rtx*. I meant to imply that I feel good about making it a mandatory dependency and expect package managers to always install it, irrespective of whether texlive is on the system -- since it is a very small dependency at half a megabyte. It is a much larger commitment to install texlive than just kpathsea. Which should return nothing if texlive isn't around.

dginev:
As we discussed with Tom in direct messages, I'm happy to merge if the default is to install using kpathsea, and the custom flag is to disable (falling back to kpsewhich, if any). So the CI file wouldn't need to change.

tkw1536:
I'll see about flipping the flag around.

tkw1536:
So turns out that rust workspaces and features are utterly broken. 
I don't think I want to spend more time into trying to fix this. 

I can do
```
cargo build -p rtx_core --no-default-features
```
to disable all default features in rtx_core. But I cannot do:
```
cargo build -p '*' --no-default-features
```
to disable default features and build all packages.

*Edit:*
All of this only works when specifying `resolver = "2"`in Cargo.toml.
I don't know enough rust to know what other side effects that potentially has.

rtx_core/src/util/pathname.rs
tkw1536: The problem is that libkpathsea is now a compile time dependency, meaning nothing works unless it is available. 
And there seems to be no easy way of obtaining it on OS X.

dginev:
Fine with me. Features are pretty scary to me, as they make it hard to reason about and maintain the code. They're useful for cross platform deployment, but given how early stage rtx is in, that can wait for a long time on my end.

My preferred approach would be to find a way to have the devel package for kpathsea work on all platforms, and have rtx agnostic to the entire question.

tkw1536:
Followup: I figured out why it is so difficult to install `kpathsea`. 

From [the homepage](https://www.tug.org/kpathsea/):
>  Kpathsea is maintained as part of the TeX Live distribution. It __has never been distributed as a standalone library__, though some distros have deployed it that way. 

Thus any installation methods are not maintained upstream, they are entirely distribution-dependent making it difficult to install from scratch outside of the few distributions that happen to maintain it.

PR #75, opened 2022-03-18
PASS t/12_grouping

dginev:
To prove my point that rtx is now unblocked, I got another test collection to pass today - `12_grouping`.

There is one notable exception that I postponed till later - related to custom binding files in the same directory as the paper. I still have to figure out how to get that ported in Rust land. So `scopemacro.tex` gets deferred in the `rtx_contrib` sub-project for now.

Grouping itself was already correct, actually. The tests were failing here due to minor markup and unrelated differences:
 - add the proper reversion call for Whatsits via `parameters.revert_arguments`. The tricky bit here was exposing a state argument to the generic `revert()` interface and then figuring out how to get the `Gullet` from there as well, at just the right times to make Rust's borrow-checker happy. Mild dancing, but in the end it was rather direct to nail down.
 - needed the ApplyNary infix rule for math grammar, so that an addition over 3 arguments produces the same XMath as latexml
 - bonus: i added the boilerplate for the `sizer` argument of constructors, but haven't implemented its internals yet.

All in all, adding another test to the list in #30


dginev:
The reason I have PRs with so many files changing is that the underlying interfaces in Rust still need to  evolve to receive the most convenient argument signatures. 

The PR sizes will start getting smaller as I get closer and closer to the coveted "parity" stage.

PR #76, opened 2022-03-27
PASS 05_tokenize

dginev:
This PR is centered around the minimal changeset for passing the missing `tokenize/mathtokens.tex`, and thus `05_tokenize`.

What was missing loosely:
 - DefRewrite, DefMathRewrite
 - DefLigature, DefMathLigature
 - XMHints, numbers, math parsing semantics for elipses, n-ary applications
 - a refactor around some hashmap getters that uses methods to check key existence without allocating the value. 
    -  This also led to an associated `0.3.1`release to the libxml wrapper

Some time was spent pondering types... some pondering remains.

What I want to do next is resync once again the test suite, since I noticed at least one file has changed ids on the perl side.

And then pick the next test case from the long list in #30


dginev:
@brucemiller one bit I noticed is that `\cdots` can no longer be `ID` on the marpa side, or we get wrongly tilted parse trees. For `a \cdot b \cdot c \cdots `, written the usual CFG way, the parse will have ` ( c * \cdots )` for an inferred invisible times between the c and the dots.

I am introducing a first new role in the PR, for now called `ELIDEOP`, for which I can design a rule that specifically expects it at the end of an infix application, as in
```rust
| term mulop tight_term elideop => infix_apply_and_elide;
```
(btw `tight_term` is a term that doesn't have any `mulop` in it yet, such as a factor chain held by invisible times)

PR #79, opened 2022-03-28
Switch to thread-safe "interior mutability" and aux State singletons

### LINKED ISSUE #78: Singleton states for convenience routines
`mouth::tokenize` and `mouth::tokenize_internal` are utility helper methods that tokenize "in vacuum", which are often helpful in avoiding binding boilerplate, for bits such as `Tokenize!("hello world")`.

As I mention in mouth.rs:
```rust
// Rust note: 1) can we avoid reinitializing a state for each tokenize call? 
// I am not sure if that is actually slow in practice, but it ought to be at least suboptimal.
// 2) If we move the $literal argument Tokenize/TokenizeInternal calls into rtx_codegen at compile_time,
// we can bunch them together as a global object in codegen maybe? Then at least we can optimize the compile pass
// + avoid runtime tokenization in the literal binding definitions.
```

It's slow enough to notice in practice on small inputs (xii.tex), where `State::new` takes 14.09% of runtime, through 195 calls to tokenize_internal.

So I think a simple well-contained issue is to keep reusing the same `STY_STATE` singleton object for these tokenization calls. It may require some mildly awkward use of `Arc<Mutex<State>` however, and that may in turn propagate to the various `Rc`s in the fields of State... Something to think about.

dginev:
This fixes #78 , via a major refactor 

1. moving away from `Rc<RefCell<_>>` interior mutability, which is not thread-safe (and is thus a little cheaper) to `Arc<RwLock<_>>` interior mutability, which *is* thread safe, and allows more versatile use in techniques such as `TokenizeInternal!`.

2. Introducing two singleton "static" State objects (in Mouth, where tokenize and tokenize_internal are defined at the moment), which get reused whenever we do tokenization "in vacuum", in particular for auxiliary binding work.

This is already felt as a speedup over both the test suite and the initially profiled file (`xii.tex`), as we save on hundreds of State inits. It's `0.03` seconds saved to be precise, but that is already *half* of the runtime for that example. The test suite fell to passing in under a second again as well, which is neat.


dginev:
An interesting side-effect of this PR is that if we *really really wanted to* we can now easily introduce a global singleton `$STATE` that is accessible from any location in the sources.

BUT, given I already did all the work needed to pass state in each function signature, and that this allows some interesting refactor possibilities for multi-thread work distribution later on, I'd rather not rush that.

PR #83, opened 2022-04-21
Optimizations in arg unpacking and cleanup

### LINKED ISSUE #45: Token::default() considered harmful
The question to answer here is whether I should erase the default token implementation, and switch to a more explicit typing interface, or keep the default method and ensure all parameter readers guard against it.

I am leaning towards entirely removing the default token, which I introduced in the first place, and does not exist in LaTeXML-proper.

The reason it was added to rtx is that certain readers in latexml make a distinction between an `undef` return value and an empty `Tokens()` value, which I wasn't prepared to respect with my type interface. 

Currently, a reader *always* returns a Tokens:
```rust
pub type ReaderFn = Fn(&mut Gullet, Vec<Option<Parameters>>, Vec<ParameterExtra>, &mut State)
    -> Result<Tokens>;
```

which makes it impossible to distinguish the undef case. In my haste, I invented a default token to add to a successful read that has no value, but that has recently bitten me in debugging during #44 .

So, I think it is time to be more correct and refactor all readers to `Option` semantics:
```rust
pub type ReaderFn = Fn(&mut Gullet, Vec<Option<Parameters>>, Vec<ParameterExtra>, &mut State)
     -> Result<Option<Tokens>>;
```

The challenge then becoming to carefully understand the difference in LaTeXML-proper, between returning an `undef`/`None` vs an empty `Tokens()/Some(Tokens!())`.

### LINKED ISSUE #65: Lighter argument unpacking in binding signatures
Currently, I had failed to avoid [cloning the argument data](https://github.com/dginev/rtx/blob/556f1be656a41ad5d27f29e4ae231a4a31da74fc/rtx_package/src/package/api_macros.rs#L347) in the binding signatures, due to having a mental block as to how to remove an element in a vector without shrinking the array and ruining the index arithmetic (in Rust).

Turns out, this is pretty obvious if one understands the `Option<T>` API. I already allowed "empty slots" in the argument list as `None` values (for things such as latex's `[]` optional arguments). To move out an existing value from an option field, the API provides the `.take()` command. So taking out the n-th argument is as simple as:

```rust
let arg = match args.get_mut(n) { // args : Vec<Option<T>>
 Some(arg) => arg.take(),
 None => None
}
```

Here is a playground link with the minimal example:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6fae5d9d9536bb203c2119882515a3bf

This (simple) trick is somewhat exciting, as it may be sufficient to eliminate the vast majority of clones left in the code. The main requirement is that the argument list must not be reusable, i.e. the closure receiving it has to consume it in its entirety - which is certainly the case for TeX bindings, as they consume the incoming Tokens arguments.

dginev:
Fixes #65 
Fixes #45 

I want to re-read the code I added to make sure it's a reasonable trade-off, but it's a step in the right direction. Rather than cloning bravely various Tokens and Digested objects (even though I made Digested fast+safe to clone by using a wrapping `Arc<>` for the inner data), the PR now detaches the individual arguments from the input Vec, or simply borrows them in the digested case (via a new `unref!` macro).

Ideally this keeps the binding execution chain a lot lighter on allocations, so we stay fast.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ A Rust port of [LaTeXML](https://github.com/brucemiller/latexml)
 
 ## Progress
 
-[![test porting progress](http://progressed.io/bar/10)](https://github.com/dginev/rtx/issues/30) | ? of ? core LaTeXML tests PASS
+[![test porting progress](http://progressed.io/bar/10)](https://github.com/dginev/rtx/issues/30) | 23 of ? core LaTeXML tests PASS
 
 ### Why?
 
--- a/README.md
+++ b/README.md
@@ -51,18 +51,19 @@ There is demonstrable need for LaTeXML in the domain of academic writing, as wel
   ```
 
 ### Fake Benchmark
- These are the times of different TeX-like engines ran over the `xii.tex` example above. That is not representative to all of TeX, but gives a good early feeling:
+ These are the times of different TeX-like engines ran over the `xii.tex` example above. That is not representative to all of TeX, but gives a minimal early feeling.
+ It will be a lot more telling to provide tikz and expl3 runtime numbers.
 
 | executable | time      |
 |------------|-----------|
 | tralics    |  0.011s   |
 | rtx        |  0.033s   |
-| tex        | 	0.067s   |
-| pdftex     |  0.125s   |
-| luatex     |  0.158s   |
-| xetex      |  0.310s   |
-| httex      |  0.333s   |
-| latexml    |  0.562s   |
+| tex        |  0.096s   |
+| pdftex     |  0.215s   |
+| luatex     |  0.226s   |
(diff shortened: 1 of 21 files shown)


dginev:
I ended up not using the `.take` approach for now, but instead directly `.remove` from the carrier Vec, when using `unpack!` and similar. This is more convenient when the arguments are _required_, as we can avoid the `Option<Tokens>` wrapper which we would get with `.take`, and have a `Tokens` directly instead.

I will likely add an `unpack_opt!` macro that allows for missing arguments. Actually yes, will add that before merging here.

dginev:
Alright, this should be enough optimizing for now - back to gaining coverage.

Issue #73, opened 2021-11-27
Memory arena for token strings


dginev:
Following [arenas in rust](https://manishearth.github.io/blog/2021/03/15/arenas-in-rust/) I read more into the elsa crate and its [FrozenIndexSet](https://docs.rs/elsa/1.4.0/elsa/index_set/struct.FrozenIndexSet.html).

It appears to be a good vehicle for allocating all raw strings used inside core conversion Token objects.

This change requires a very large refactor related to lifetimes, essentially adding a lifetime to any piece of data that directly or indirectly may contain a Token. But what we get in return is valuable - and a lot closer to the benefits of Knuth's original implementation.

In essence any raw string used as the contents of a token will only need to be allocated once, and then be referred to immutably for the rest of the conversion process. Compare this to latexml's perl implementation, where each time we see a `T_CS("\relax")` we allocate a new string for `"\relax"`. In latexml's defense a partial optimization exists, where some standard tokens (`T_BEGIN`, `T_END`, `T_MATH`, etc) are allocated once as constants. But since it's a partial treatment, it can lead to some trickiness... 

Avoiding the allocations has even more benefits than the obvious ones in Token itself. When binding code is run, it often creates tokens from within the rust closure that implements some XML-friendly behavior. Creating new Token objects is fine and good for that, but reallocating the strings has a visible impact when looking at performance benchmarking. There are a lot of low-level clone and drop calls related to string management.

Lastly, managing immutable pointers to the raw string allows for a cleaner Token abstraction, where we can fully see them as frozen immutable objects that contain a catcode enum and a string pointer, and auto-derive the `Copy` trait for lightweight duplication.

The refactor is painful, but I think worth trying, since the performance upgrade ought to be rather substantial.

dginev:
I think this will *also* allow lightweight extensions down the line. Adding two extra fields to `Token`, say `line: u16` and `column: u16` in order to bookkeep an origin locator on each token, while also adding a `source_file: &'origin str` will still allow to retain the auto-derived `Copy` trait, and avoid any heavy clone and drop operations. 

And if a single source file contains thousands of tokens, we'll only ever allocate the filename string once, and bookkeep a pointer to that string in each Token, rather than allocate and drop thousands of times.

dginev:
[...]

PR #85, opened 2022-04-28
Re-sync to pass structure/itemize.tex

dginev:
Notable upgrades:
 * Tokens arguments are now passed inside an `Option<Tokens>` wrapper, so that we have a natural way to model optional arguments, such as `OptionalUndigested` and friends.
     *  In these cases, there is now a direct equivalent to a perl `if (!defined $tokens)`, which is the rust `if tokens.is_none()`.
 * A lot of the code that dealt with `CC_ARG` and `packParameters` had gotten upgraded since I first ported, so I updated most of those pieces, to get the right argument passing again.
 * Similarly, I had to be very very precise with `RefStepItemCounter` and `RefStepId`, to make sure the counter acrobatics matches exactly.

What I gain from all of that, is that `structure/itemize.tex` passes fully again. Whew. 

A bit too much work for one test, I hope I don't have to keep making this comment...


dginev:
Another motivating comment, here are the times running the now passing test, via e.g.
```
$ cargo build --release
$ time ./target/release/rtx rtx/tests/structure/itemize.tex
```

cmd | seconds
:---- | ----:
tralics | 0.010
pandoc | 0.046
hevea | 0.055
**rtx** | **0.081**
latex | 0.165
pdflatex | 0.245
latex2html | 0.316
plastex | 0.512
htlatex | 0.905
**latexml** | **2.177**
lwarp | 2.442

I have to say I am highly curious about what magic tralics is using to stay at near-constant speeds. I can only guess that they're avoiding heap allocations to a maximum, as the original TeX itself does. When I used valgrind to benchmark the execution of rtx, a good chunk of time is spent on allocating string names, hashes, vectors and initializing the indirect model and marpa grammar.
These are all heap-heavy operations.

I think the goal was never tralics, so this is more a curiosity than a real concern, but it's good to know one competitor has really been optimized for performance.  rtx is about 27 times faster than latexml on this small test (for now).

dginev:
Here is a screenshot of said performance analysis, ordered by "exclusive" i.e. "Self" time, of specific subroutines.

We can see:
- 9+3=12% of the time rtx is freeing memory from the heap via free and int_free.
- 8.8% of the time rtx is allocating new strings via the `write!` macro,
- 8.6% of the time is spent at the low-level hashmap allocations, 
- another 6.1+5.1=12% of the time is spent allocating memory to the heap via malloc() and int_malloc().

That is already 40% of all runtime! And I haven't even counted all other auxiliary calls that have to deal with heap memory management, such as building and rebuilding hashes and vectors. I think this confirms the hypothesis that the main remaining differentiator between tralics and rtx are the degrees of using of heap memory.

That said, the latexml algorithms are very much heap-hungry, as perl allocates *all* of its data on the heap. I would need to reinvent entire high-level abstractions to move away from that, and I am not sure it is relevant before the first major rewrite is complete. Good to know in advance though!

<details>

image

</details>

dginev:
Update: Indeed, tralics follows Knuth's approach to embrace opcodes and design(+maintain) a huge enum for all known commands:

https://github.com/tralics/tralics/blob/d1a04d428c0377c02912bf1392d8be837db0be32/src/tralics/enums.h#L162-L173

Also the tralics state is a [simple static hashmap](https://github.com/tralics/tralics/blob/d1a04d428c0377c02912bf1392d8be837db0be32/src/Symcode.cpp#L3-L7) without any scoping infrastructure _at all_, that connects the symcode numbers for each command to its "action". Indeed quite close to the TeXbook.

If I am understanding currectly, the speedy performance is also backed by a provision of token constants which also won't get allocated during runtime:

https://github.com/tralics/tralics/blob/e11a8f05751c27e751a522abd40f6f055365b66a/src/tralics/Hashtab.h#L33-L51

And possibly there are a handful of extra tricks I haven't seen yet. (for one example, I saw a lot of data structures, such as TokenList, get modified in place whenever possible, to minimize runtime cost).

Who knew :>

dginev:
I am tempted to conclude the tralics investigation with - "good to know!" and a brainstormy note.

Taking the direction of #73 , one could imagine allocating each token exactly once, and pointing into a global lookup table, which is allocated at compile time via `enum`s for the standard tokens, exactly as in tralics and TeX's B book. That would be a major refactor of what a `Token` is during classic latexml processing however, and it needs a lot of fiddling to get both the modern ergonomics, together with the allocation-free behavior. And it would open a series of questions about attaching Locator objects to Tokens, which bring along their own allocations - and require a real "struct" for the Token.

I think there is a natural path to a middle-ground approach, in fact one quite close to #73, where we amortize the allocations for token strings.

PR #95, opened 2023-01-24
Unified type interface for binding arguments, Part 1

dginev:
This PR marks a "part 1" of the refactoring that gets us to typed binding closures.

Namely, an introduction to a badly named (needs a better name!) `ArgWrap` type, which wraps around any argument of a binding - commonly things like `Tokens`, `Number`, single `Token`, `Dimension`, etc. Or the advanced and unusual `RegisterDefinition` trick from the `Register` DefParameterType.

A lot is left to be desired, but I would really want to move this into master, so that I can align to the latest rust clippy lints and keep the next PR small.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,3 +69,10 @@ Requires Rust `stable` v1.32, and newer.
     ```bash
     $ cargo run --release --bin rtx rtx/tests/hello/hello.tex
     ```
+
+### Docs
+
+To generate the project documentation locally, run:
+```bash
+$ cargo doc --workspace --no-deps --open
+```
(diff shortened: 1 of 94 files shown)


dginev:
The most important conceptual piece of this PR is the high-level change of the binding closures from concrete Tokens arguments, to "wrapped" arguments. Here is a screenshot:

image

The rest, sadly, were mostly low-level details and wrestling the Rust boilerplate.

I am still writing very boilerplaty code, which should eventually get compressed back down through smarter dependency management and use of generics.

PR #97, opened 2023-01-24
Typed bindings, part 2

dginev:
Now that we have some initial typed interface in place, I have a first demo of a typed binding.
This PR records that state - the next and "near final" step will be to make *all* bindings compile into such concretely typed closures.

Here are some screenshots:

A draft typed binding signature may look close to:
```rust
  TypedMacro!("\\sampler" number:Number, token:Token, dimension:Dimension => sub[gullet,state] {
    // ...
  });
```
Eventually this will be an intermediate step of expanding a `DefMacro!`.
At the final expansion step, there is a type cast done in a macro behind the scenes, of the following flavor:
```rust
let $var: $ptype = match args.remove(0).try_into() {
  Ok(v) => v,
  Err(e) => {
    Error!("expected", "argument", $gullet, None, e);
    $ptype::default()
  }
};
```
if the concrete `$ptype` here was say a `Number`, this would require a directive for the generic `TryFrom` trait over the wrapping type `ArgWrap` in the argument module:
```rust
impl TryFrom<ArgWrap> for Number {
  type Error = crate::common::error::Error;
  fn try_from(aw: ArgWrap) -> Result<Number> { aw.try_to_number() }
}
```
where we can define and reuse a generally handy `try_to_number` method for ArgWrap.

---
Once all of that is in place, we can actually write bindings with well-typed arguments!

My VSCode `rust-analyzer` plugin can now autocomplete the `Number` methods, as it recognizes a variable of that type:

similarly if instead the variable was the second `Token` argument:

In closures of this nature, Rust's compiler will protect a binding author from any misconceptions about the types of the data passed in, avoiding unfortunate calls that have traditionally lead to perl "die"s.

Moving on to the next step...


dginev:
Ok. I have gotten the PR to a comfortable state where I can run "cargo expand" and find informative differences between `TypedMacro!` and `DefMacro!`, namely:
```
cd rtx_contrib/ ; cargo expand typed_bindings_sty 
```

The next component to adapt is the `parse_prototype` function, which currently executes at runtime -- it needs to be submerged as a compilation step in the `_codegen` subcrate. I would find it cleaner to do that in a separate PR, so I'll merge here and start a new branch.

PR #98, opened 2023-01-26
[MVP] Compile-time expansion of string prototypes in bindings

dginev:
Now we're getting somewhere! This is pretty neat:

```rust
  TypedMacro!("\\sampler Number Token Dimension", sub[gullet, (number, token, dimension), _state] {
    number.value_of(); // OK
  });
```

but if we change the numeric argument to Token in the string(!!) prototype and simply save + reanalyze the file (again, vscode with rust-analyzer here), we see:

This was accomplished with a little extra codegen tedium, which was mostly straightforward since we already had the `parse_prototype` function implemented. Namely:
 - submerge `parse_prototype` from rtx_package down into rtx_core, making it available to rtx_codegen
 - a minor refactor that allows parsing prototypes without access to state (i.e. without initializing the parameters), which is great for compile-time work.
 - passing the parameter names (Number, Token, Dimension here) back out as type identifiers via the `format_ident!` macro of the quote crate, making them available as types during runtime.

Various boilerplate can still be improved on, but I am quite happy that I have a working demo to showcase today.


dginev:
@tkw1536 and @brucemiller if you are interested how the code switch looks to move from the `let mut parameters` snippet we looked at yesterday into an immutable version of it, here it is isolated in a single commit:

a3d868770cb20cd375827575205d48131485d2b0

Rust makes these kinds of code polish tweaks nice and easy 👍🏻 
Just a simple [macro repeat](https://doc.rust-lang.org/rust-by-example/macros/repeat.html)

dginev:
Another note is that @tkw1536 is onto something with his mention of using arrays instead of a dynamic vector for the case of parameters. TeX has a maximum of 9 parameters, so we can quite safely use a 9-element Array, wrapped with a bit of metadata for how many arguments are actually there.

It turns out, Rust's `Box` model allows for moving arrays, which are by default in stack memory, onto heap memory ([see here](https://stackoverflow.com/questions/25805174/creating-a-fixed-size-array-on-heap-in-rust/68122278#68122278) ).

And indeed there are helper libraries that have implemented exactly that, such as the `ArrayVec` type in tinyvec:
https://docs.rs/tinyvec/latest/tinyvec/struct.ArrayVec.html

Then I can change my macros to use their "draining iterator" and call `.next()`, instead of the rather brutish `.remove(0)` that I keep calling on the Vec data type.

It's Rust - if you want a performance optimization somewhere, it's possible.

dginev:
Alright, I have integrated the first prototype as a DefMacro variant, and I think this PR can now be merged. The typed interface is still _experimental_ and needs more boilerplate code to be finished for 1) all parameter types, 2) all Def* variations. Some of these are harder than others (e.g. parameters with default values in the replacement, parameters where the name does not match the arg type (e.g. `Until:XXX` -> `Tokens`), and parameters that get completely Skipped.

But now that the infrastructure is there for _one_ variation, the rest are just elbow grease.

The working example is now:
```rust
  DefMacro!("\\sampler Number Token Dimension",
    sub[gullet, (number, token, dimension), state] {
      ...
  });
```

Issue #101, opened 2023-01-28
Binary speed+size for releases


dginev:
An interesting observation while catching up on some [rust materials](https://www.youtube.com/watch?v=b2qe3L4BX-Y) - the exact cargo flags used during the builds have a significant impact on the final binary's size (and maybe speed?)

The current setup uses the default `cargo test --release` to produce an `rtx` executable in `./target/release/rtx` which is 20MB in size.

If instead we add to Cargo.toml:
```toml
[profile.release]
strip = true
lto = true
codegen-units = 1
```

The rtx size falls down to 11 MB when rebuilding.

A related detail is to consider the bloat of the project and its dependencies (20 MB variant):

<details>

```
$ cargo bloat --release --crates
Compiling ...
Analyzing target/release/rtx

 File  .text     Size Name
20.7%  56.6%   4.1MiB rtx_package
 4.9%  13.5% 991.3KiB std
 4.7%  12.9% 950.9KiB rtx_core
 1.8%   5.0% 368.5KiB rtx
 1.0%   2.7% 200.1KiB regex
 0.9%   2.5% 185.7KiB regex_syntax
 0.8%   2.3% 165.5KiB rtx_math_parser
 0.5%   1.3%  95.7KiB libmarpa_sys
 0.3%   0.9%  69.2KiB aho_corasick
 0.2%   0.5%  38.5KiB rtx_contrib
 0.1%   0.4%  30.0KiB chrono
 0.1%   0.3%  25.1KiB libxml
 0.1%   0.3%  19.4KiB marpa
 0.1%   0.2%  12.4KiB memchr
 0.1%   0.2%  11.6KiB [Unknown]
 0.1%   0.1%  10.9KiB miniz_oxide
 0.0%   0.1%   3.9KiB kpathsea
 0.0%   0.0%   3.2KiB ansi_term
 0.0%   0.0%   1.9KiB iana_time_zone
 0.0%   0.0%   1.6KiB which
36.6% 100.0%   7.2MiB .text section size, the file size is 19.6MiB
```

own:
```
$ cargo bloat --release 
Compiling ...
Analyzing target/release/rtx

 File  .text     Size       Crate Name
23.9%  65.3%   4.7MiB             [10722 Others]
 4.7%  12.9% 948.0KiB rtx_package rtx_package::package::pool::tex_appendix_b_p358::load_definitions
[...]

PR #106, opened 2023-02-15
PASS expansion/{noexpand, noexpand_conditional}

dginev:
Passing two more tests from 01_expansion.

Upgrades related to:
 - accuracy of the noexpand directives
 - managing fonts when inserting math tokens in scripts

I just barely didn't drown with the Font-related fixes. My setup for fonts is very unpleasant, and needs a rethink. In this PR I switched to using the `_font` attribute, the way latexml does, but I have not standardized that treatment enough. I am also not sure it's the best approach, but it is so hard (for me) to track the way latexml deals with the `_font` and `font` attributes/options in the various Document methods, that it seems most predictable to just follow along the exact logic seen in perl.


dginev:
There are also a variety of changes here that are leftover from two experiments I kind-of succeeded at, but eventually backed out of:
 - A globally available singleton `STATE` that was realized via `Arc<RwLock<StateData>>` in a `lazy_static!`, with runtime read/write locks
 - An alternative to `PartialEq` called `StatefulEq` which tested equality with an additional `&State` argument, in order to consult deeper relationships, such us whether the `meanings` of two tokens are identical (via `STATE.lookupMeaning`)

Both of those were not _actually_ needed (instead, I needed to flesh out more of the Register details), so I backed away most of the code that enabled them, but not all of it.

dginev:
Oh, I also made a dedicated `def_math` function, as mentioned in #101 . 

That got the smallest `rtx` binary I can build down to `8.2MB` from `8.8MB`. Neat.

Will merge here and move on...

PR #112, opened 2023-02-25
PASS expansion/lettercase

dginev:
Added a variety of unrelated boilerplate here + a `cargo fmt`, so this PR is a little hard to read.

The key changes relate to `\uppercase` and `\lowercase`, and the infrastructure they use behind the scenes.

And despite the large diffs, most of the changes were quick to make - 90% of the time for this PR went into painfully tracing and debugging the full call stack behind the `\MakeUppercase` macro, until I bottomed out at realizing that I had declared a `DefMacro!` instead of a `DefPrimitive!` for `\uppercase` which completely jumbled up the expansion stack.

I added a small experiment trying to avoid spending time on allocating empty return `Tokens` objects, with a new `NO_TOKENS` constant, will have to do some measurements to see if that holds up.

Tests (`cargo test --release --tests`) now take 1.6 seconds on my machine , up from 1.4 seconds. I'm thinking of doing a more thorough performance profile once I flesh out enough tests for 2 seconds runtime.


dginev:
Oh, and I should mention that I changed the `lettercase` test to use commas instead of a table, so that I can pass it without implementing `{tabular}` and alignment - which are still missing from the Rust port.

PR #119, opened 2023-03-01
PASS expansion/testmultido

dginev:
Still working here, but wanted to file the PR early and send a signal flare to @brucemiller .

I will need some hand-holding to try and disentangle the "inner" and "extra" fields associated with a `Parameter`, as well as the exact types of data (and maybe timing) that flow through them. Also for their method calls of `reparseArgument` for parameters such as `{Number}` or `{Float}`, which is what this particular test exercises.

There were 3 big missing pieces for the multido.sty test to work: 
- Writing the actual Rust code for the Perl .ltxml binding
- Realizing that `{Float}` isn't yet handled correctly, due to the Parameter `extra` logic not being fully ported.
- Also realizing that `fixpoint()` can and *does* overflow when large SP values are passed in with the sp unit (which seemingly implies an innocuous `1.0` multiplier). But if one multiplies a large SP value stored in an `f32` by the 65536 UNITY, there is an overflow when casting back to `i32` with kround. In the multido test, one sp value I was seeing was `131072`, so the final multiplication produced `8,589,934,592`, which is significantly greater than the i32::MAX of `2,147,483,647`.

A side piece, which may or may not be worth bringing up now, is that the checked in PDF for `testmultido.pdf` doesn't _exactly_ match the XML. Consider for instance the first paragraph's numeric lines, each of which starting with a 2. In the PDF, the precision shown is:
```
2, ...
2, ...
2.0, ...
2.00, ...
```

while in the XML I see
```
2.0, ...
2, ...
2.0, ...
2.0, ...
```

Glad to be in the clear where I know exactly what was missing for the test to pass. I'll tinker a little more, but I need to better understand the abstractions to actually flesh out the code to a decent state of affairs. It's rather messy at the moment.


dginev:
Another exciting low-level Rust problem is dealing with `f32` precision. In this test, one can also witness the following exciting result:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e539068b30a0b724b5795ac9466bb377

```
-4.1 + -3.05 = -7.1499996
```

Our test expects `-7.15`, which is both what pdflatex and Perl output. I'll have to get a hang on the exact choices Rust has made with `f32`, and then figure out the mitigation strategy...

Edit: if anyone is wondering what result one gets with `f64`, it is `-7.1499999999999995`

brucemiller:
Nope, you don't want to do floating point at all.
You need to duplicate TeX's fixed point arithmetic
(using an integer internally).

dginev:
@brucemiller right - I have ported that in the places where latexml already has taken that path.

The values I mentioned above come from adding together two `Float` objects (as in `LaTeXML::Common::Float`), which - I think - is using the native Perl floats and Perl numeric operations on floats. But maybe your point is that this should be done entirely different in Rust, never touching `f32`.

brucemiller:
hmm... Common::Float is kinda illegit anyway :>

dginev:
Aha! You scared me so I got an adrenaline rush and solved my mistake. I had to use `floatformat()` in the implementation of `to_string()` for the `Float` objects.

Now it matches. Hopefully a huge refactor was just avoided (fingers crossed)

dginev:
The multido test now passes matching exactly the XML produced by mainline latexml, so I think I can merge here.

It is still the case that the `Paramater` extra/inner details are rather hacky and need to be polished, but that can be done in a follow-up PR.

PR #121, opened 2023-03-03
PASS expansion/whichinput

dginev:
A number of open questions are still lurking here, but `expansion/whichinput` now passes.

The `find_file` logic and `input_definitions` logic still have a lot of fleshing out left.

It is also now the case that the `kpsewhich` searches are done more consistently, which appears to have single-handedly doubled the test times (from 1.6s now at 3.1s). This may go away once I handle the "search paths" correctly, which also needs a bit more work.

And also, I haven't added the cache around the `kpsewhich` calls. Maybe I should do that before merging here, since I am already seeing multiple kpse calls for the same filename.


dginev:
Fun fact, the cache building logic in Pathname.pm takes `0.8 seconds` to complete.
```
$ time perl -MLaTeXML::Util::Pathname -e 'LaTeXML::Util::Pathname::build_kpse_cache();'

real	0m0.818s
user	0m0.784s
sys	0m0.034s
```

Let's see if there is a way to improve on that...

dginev:
I may be able to avoid re-implementing a KPSE cache in rtx, since we can simply reuse the same initialized low-level `kpathsea_new()` object through a lazy_static `Arc<Mutex<Kpse>>` singleton.

This lowers the test runtime on my machine back down to 1.9 seconds, which is good enough to merge.

PR #123, opened 2023-03-08
PASS expansion/urls

dginev:
Since the `url.sty` binding has been in place for some time already, not much was left to do to pass the essence of this test.

Except that the test also exercises `URLs in footnotes`, and I had no implementation for `\footnote`.

So this PR adds the basic footnote-related functionality, together with the basic Document infrastructure that drives their relocation.

More remains to be fleshed out, but it suffices to pass this particular test...


dginev:
Motivating aside: The release binary for rtx finishes the `expansion/urls.tex` test in `0.075s`, while latexml completes it in `2.021s` on my machine.

Issue #127, opened 2023-03-13
are we comitted to 64 bit numbers?


dginev:
Here is a recent screenshot from debugging the etoolbox test, for something as simple as a dimension value of `10in`:

![Screenshot 2023-03-13 at 09-27-42 Rust Playground](https://user-images.githubusercontent.com/348975/224715920-7540aa77-9faf-462f-ac99-5287909716c8.png)

Two aspects seem clear from my test run:
 - perl is using `f64` floats, at least on my machines (which use 64-bit architectures)
 - The algorithmic reliance on the `65536` multiplier will keep causing various overflows for 32-bit numbers.

For the first milestone, where I am hoping to gain parity with the Perl codebase, I will change to using `f64` internally. But maybe we want to reconsider some of the techniques long term? Unsure.

dginev:
I should have mentioned, the screenshot is from the Rust definition of an inch:
```rust
pub static ref UNITS: HashMap<String, f32> = map!(
 "in" => 72.27 * 65536.0,   //...
```
which tracks the same definition in LaTeXML's State.pm:
```perl
$$self{value}{UNITS} = [{
  in => 72.27 * 65536, #...
```

brucemiller:
TeX uses fixed point numbers. An inch to LaTeXML is intended to be 
exactly the same approximation as it is in TeX.

dginev:
Here is a quote from the B Book:
![image](https://user-images.githubusercontent.com/348975/224761980-5d8d618f-ccbf-4196-bff7-68a94f609452.png)

And here is the Rust inaccuracy, if the _documented_ number is realized directly as `f32`:

![image](https://user-images.githubusercontent.com/348975/224761685-23b850a1-1fc3-4571-910a-f16a2aa0c903.png)

To my understanding, we have a choice between 
  1. using `f64`, as in Perl, or
  2. not representing inches directly, but following the B Book more strictly, and implementing the `set_conversion` logic as presented in 458.:

![image](https://user-images.githubusercontent.com/348975/224762796-47b1b714-bf4b-4270-bcf5-27fcfb4142de.png)

brucemiller:
2. Fixed point is an integer with assumed position of decimal.  Sloppily left as float in State's UNITS, but always used with the appropriate adjustments in Dimensions, etc.

The relevant test is what you get with `\dim=1in \the\dim`

PR #128, opened 2023-03-14
PASS expansion/etoolbox

dginev:
This PR concludes the `expansion` tests - which now ALL pass.

I left etoolbox for last, since I expected a lot of work to go into it (and it did).
Here is a summarized list:
 - add etoolbox binding
 - track down `f32` overflows and switch to `f64` to match the Perl implementation (and add long-term tracking issue with #127 )
 - flesh out `-h@@k` machinery
 - `ExpansionBody::Closure` can now serialize the raw pointer of the closure, in the same way we had `CODE(0x5....)` addresses in Perl. Moreover, if that string is taken from the `\meaning` parser and tokenized+set as an Expandable definition, testing for equality the resulting `ExpansionBody::Tokens` with the original `ExpansionBody::Closure` will only succeed if the two represent the same exact closure. Quite important for the etoolbox tests.
 - localized better the `CURRENT_TOKEN` set/get/expire logic
 - tracked down and corrected the subtleties around `is_dont_expandable` (hardest piece to debug)
 - caught a few other minor porting inaccuracies

Running the full `01_expansion` suite currently takes exactly a second on 1 CPU, compared to 20.5s in Perl.


dginev:
Also marked the milestone with a [0.3.0](https://github.com/dginev/rtx/releases/tag/0.3.0) release.

PR #130, opened 2023-03-14
PASS digestion/chardefs

dginev:
This PR:
- add `Font::decode_fontname`
- add font closures for DefConstructor
- speedup for `lookup_catcode` by avoiding memory allocation

And a basic `\mathchardef` works!


dginev:
* also passes digestion/testctr

Issue #82, opened 2022-04-18
Manually copy over perldoc as rustdoc


dginev:
This will be quite a lot of copy-pasting and proofreading the resulting doc pages, likely best done when immediately prior switching the repository to publicly visible.

The idea is that most methods in latexml have a perldoc entry at the end of their `.pm` file, which need to be turned into the triple-slash comments expected by rustdoc, just before the method declaration.

As a random example, take State.pm's `deactivateScope`, which has a perldoc entry that looks like:

```perldoc
=item C<< $STATE->deactivateScope($scope); >>

Removes any definitions that were associated with the named C<$scope>.
Normally not needed, since a scopes definitions are locally bound anyway.

```

This would need to be moved over in `rtx/rtx_core/src/state.rs` as:
```rust
/// Removes any definitions that were associated with the named `scope`.
/// Normally not needed, since a scopes definitions are locally bound anyway.
pub fn deactivate_scope(&mut self, scope: &str) {
```

Wash, rinse, repeat, for the entire code base.

dginev:
I have started working on documentation. 

The branch I'm working on is still mostly empty, but I have configured it to generate gh-pages documentation at:

https://dginev.github.io/latexml-oxide/latexml/

Working specifically on `rtx_core` for starters. It's helpful to look at the gh-pages, so that I can adjust the formatting for any doc strings that don't fit naturally.

dginev:
This is mostly resolved for now, pieces have moved sufficiently to make it difficult to do 1-to-1

PR #135, opened 2023-03-28
Docs sprint, Q1 2023, phase 1

dginev:
This is phase 1 of the documentation sprint for #82. It started with the simple addition of triple slash doc strings to functions as in 
```rust
/// how this works
fn function_name(...) { ... }
```

In the process I got to do a variety of small/inconsequential refactors such as removing unused arguments in a couple of functions, or even entirely removing a memory-guarding type (goodbye `RegisterCell`). Which in turn led to dropping (or tightening) some of the compiler silencing I had done with `#[allow(unused_mut, dead_code, unused_variables)`.

As a high-level overview summary:
1. dropped directives from rtx_core/lib.rs:
    ```rust
    #![allow(dead_code, unused_variables, unused_mut, unused_macros, clippy::trivial_regex)]
    ```
2. reduced directives from rtx_package/lib.rs, from:
    ```rust
    #![allow(dead_code, unused_variables, unused_mut, unused_macros)]
    #![allow(clippy::unused_unit, clippy::implicit_hasher, clippy::trivial_regex)]
    ```
    to:
    ```rust
    #![allow(clippy::implicit_hasher, clippy::trivial_regex)]
    ```

cargo build and clippy now raise a lot more warnings for bad style code.

There is more of that to tease away, and over 1,200 undocumented entries to still add doc strings for, but I think this is already self-contained enough to mint one PR. I will keep adding the changes with these "phase `i`" markers in the title.


dginev:
This PR also adds an automatic gh-pages build script for updating the documentation, which should be visible at:
http://dginev.github.io/rtx/rtx/index.html

I don't quite like how the top-level entry page looks, likely will need some more reorganizing. 
The rtx_core page looks a little better:
http://dginev.github.io/rtx/rtx_core/index.html

PR #136, opened 2023-03-30
switch to FxHasher from std HashMap

dginev:
I was taking a break from documentation, and played around with benchmarking a file I hadn't before (the expansion/etex test). I noticed a high footprint (relative to how fast everything is) of the `core::hash::BuildHasher::hash_one` calls, which are used whenever we interface with the HashMap contents (e.g. in the usual lookups `hashmap.get(key)`). 
We use a *lot* of hashmap access, given that 
1. almost all data management happens in the State tables and 
2. we're coming from a Perl context, where most structured data is a hashmap or an array

So, I searched around for some discussion on the Rust hashing, and indeed there is quite a bit written. I quite enjoyed "[A brutally effective hash function in Rust](https://nnethercote.github.io/2021/12/08/a-brutally-effective-hash-function-in-rust.html)", which went into some detail about which trade-offs are chosen and why.

Since latexml isn't interfacing with any security-critical applications, and we are never intentionally exposing the internal hashes to the world, there was a good case to be made that the default `HashMap` isn't the perfect trade-off for `State` - it does too much in trying to remain as secure and HashDoS-resilient as possible. In turn, `FxHasher` tries to stay as quick as possible by doing as little as possible - and offers no extra guarantees. I took 30 minutes and refactored into using `FxHashMap` and `FxHashSet`, then compared a simple test run:

```
```

<table><thead>
<th>rustc_hash</th><th>std::collections</th>
</thead><tbody>
<tr><td>

```
real	0m2.026s
user	0m1.798s
sys	0m0.226s
```

</td><td>

```
real	0m2.309s
user	0m2.082s
sys	0m0.229s
```

</td>
</tr></tbody></table>

And wow, a 10% speedup! 

Analyzing some perf visualizations confirms this, so I'm tentatively opening a PR. It would be nice to study a big test in some extra detail (e.g. the 3000 primes perf charts), but - in theory - FxHasher intentionally does less work than the standard rust HashMap hasher, so the speedup ought to be visible in various other tests.

It would be a rather nice win if 10% can be gained so easily, with this much infrastructure already fleshed out.

### DIFF
--- a/rtx/Cargo.toml
+++ b/rtx/Cargo.toml
@@ -30,6 +30,7 @@ libxml = "0.3.2"
 lazy_static = "1.0"
 regex = "1.7.1"
 glob = "0.3"
+rustc-hash = "1.1.0"
 rtx_codegen = {path = "../rtx_codegen"}
 rtx_core = {path = "../rtx_core"}
 rtx_package = {path = "../rtx_package"}
--- a/rtx/src/core.rs
+++ b/rtx/src/core.rs
@@ -3,6 +3,7 @@ use regex::{Captures, Regex};
 use std::borrow::Cow;
 use std::path::Path;
 use std::sync::Arc;
+use rustc_hash::{FxHashMap as HashMap};
 
 use rtx_core::common::error::{note_begin, note_end, Result};
 use rtx_core::common::DigestionMode;
--- a/rtx/src/util/test.rs
+++ b/rtx/src/util/test.rs
@@ -2,7 +2,7 @@ use glob::glob;
 use libxml::parser::Parser;
 use libxml::tree::Document as XmlDoc;
 use libxml::tree::{Node, SaveOptions};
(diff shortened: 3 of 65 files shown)


dginev:
examining a valgrind benchmark of converting `expansion/etex` this morning, which confirms the observations. 


Here's a screenshot. left side is `std::collections`, right side are the newly added `rustc_hash` collections (sorted by exclusive "Self" time):

<details>

image

</details>

The total "instruction fetch cost" is 186 million for the `std` variant, and 154 million for this PR's `rustc_hash`-based variant. That is a 17% decrease in IR fetch - nice.

dginev:
We discussed this on the call today a little, and decided to give the transition a try. Some upgrades to the code quality were brought up as well (e.g. using a compiler flag to choose the hasher), which I may add a littler later.

PR #139, opened 2023-04-08
string-interning arena (Tokens, Boxes, State and Model)

dginev:
One way of fixing #73 .

Interestingly enough, `string-interner` offered a simple enough interface and the refactor was just the kind of task one can calmly walk through on a Saturday afternoon.

I suspect the benefits would be felt even more on larger documents, but profiling with this PR shows another 1% of instruction fetches has been avoided.

I think I saw less of a speedup than I expected due to already having used `Cow` in tokens, which had avoided the overhead for all static string values. And in that case *new* overhead got introduced, as getting the text of the token now requires an arena lookup.

The worst aspect of this approach is that I ended up quickly using a mutable static singleton, which ties the entire project to a single thread. For a multi-threaded arena I would need to add a `Mutex` lock (which will a bit of extra overhead) or move to a different provided for interning such as `lasso`.

Edit: There is also an independent refactor moving away from `lazy_static` and towards the `once_cell::sync::Lazy` alternative, as it may soon become standardized in Rust's `std` library. Should make the "official" refactor to follow easier.


dginev:
1. I have also added string interning for the Model and IndirectModel, which were the next biggest allocators in the little test I was benchmarking.

2. I've also moved the interner to use FxHasher, which is a smidge faster.

It's _marginally_ faster, but where the difference is really impressive is the "Instruction Fetch" count in kcachegrind. For `tests/structure/itemize.tex` the changes in the PR got us down from 258 million IF to 225 million IF. About 13% less.

This is quite the low-level tweaking, and I am mildly worried I may be over-interpreting hardware artifacts from my own individual machine, but the general concept seems sound enough - if we intern the "hot path" of string use in a single global arena, we can avoid thousands (if not millions) of allocations on the heap, and we are also conceptually closer to the way TeX/tralics deal with Tokens (unique numeric array indexes for the standard collection).

Will leave the PR stewing for a little, but I'm currently closer to merging it and moving forward. It's not too painful to undo later, though it will take some work :)

dginev:
The Token struct is now:
```rust
pub struct Token {
  /// an arena id for the character content for this token
  pub text: SymbolU32,
  /// a TeX catcode
  pub code: Catcode,
  /// possibly smuggled inner token (for \noexpand)
  pub smuggled: Option<Box<Token>>,
}
```

Where `SymbolU32` is a unique id for the interned textual content. That ticket is issued via `arena::pin(text)` and the content can be re-obtained by calling `arena::resolve(ticket)`, which returns a static reference to the internet text. Quite handy, as I have managed to avoid a number of String allocations in the Document methods already, as we have more opportunity to deal with string slices (`&'static str`) when the tickets get resolved.

The PR is using these `SymbolU32` for Token, Tbox and Model, at the moment. Can gradually expand to other hot allocation paths, if/when observed.

dginev:
Update:
 - checked the `ustr` crate as an alternative. A medium-sized refactor later I didn't see performance improve. Also, the ustr interface is more restrictive, so it's harder to use. The use of `AsRef<str>` by string-interner makes it quite convenient for using differently typed arguments in interfaces such as `T_CS!()`
 - added a refactor that now relies on SymbolU32 for the State tables as well. Note that that now imposes a hard size limit on the possible entries in the value table, at the max element of u32 (the exact number is 4,294,967,295) - or we'll start seeing collisions.
 - This finally surfaced a threading failure in the tests, as the unit tests for State started experiencing data races. So, the time has come - I added a `Mutex` guard to `arena::pin`. It is now thread-safe in one of the most simplistic ways possible. A migration to `lasso` continues to be a possibility one day...

In case anyone is wondering - a small test such as `structure/itemize.tex`, with the limited implementation of the pool files, allocates 3,200 unique strings in the arena.

dginev:
A fresh comparison with master shows about 10-15% speedup of this branch with master.

Tempted to merge today and start working on alignments (finally!)

Issue #140, opened 2023-04-14
thead_local! or no memory arena


dginev:
A follow-up to #73 after a very useful round of feedback from @tkw1536 and @brucemiller 

It turns out I had misunderstood exactly what I was seeing in the current test runs, as I wasn't stress-testing the right tests with a sufficiently high number of threads. A properly multi-threaded test run is now **slow** (as long as 11 seconds), while a single-threaded run is 2.4 seconds.

The reason is as Tom predicted - the Mutex locks lead to long wait times.

I think it is reasonable to provide a thread-local memory Arena, just as we provide a thread-local State. The tricky thing is making it work correctly with Rust's lifetimes. The easy part are the ticket integers - they are Copy types that always have ownership. But any `&'arena str` reference that is taken through `arena::resolve(ticket)` needs to have a very tight lifetime - so it must be used up fully before the thread lifetime expires.

It is a bit scary to try and implement that, since there is a risk of "lifetime creep" where the method signatures get quite complex to read and reason about. But maybe it is worth the trouble - if we had a massively parallel web service that ran a standalone conversion process on each *thread*, having a thread-local State and memory arena would be highly beneficial for throughput. 

So I should at least try before I give up on the arena idea completely (which I think is the manageable alternative, now that I've seen the starvation issues).

dginev:
Sometimes I am too stubborn in admitting that I've made a mistake... 
I spent a full day of typing today to create a `thread_local!` variant on the string interner approach, where each thread gets a local memory arena.

This achieves *no* lock waiting (or data racing) between threads.

But it also introduces the potential for deadlocks within a *single* thread! Which is indeed where I left the work. There is a [thread-local-nightmare](https://github.com/dginev/rtx/commit/64ecc94af11d8c7caf843aae205238ecdc3640ed) branch commit that contains the entire expedition.

In brief, ensuring that we are in the local lifetime of a current thread switches the well-known cs name getter of Token:
```rust
  let cs_name : Cow<&'static, str> = cs.get_cs_name();
  // code using cs_name
```

to an indirect closure-based handler, which holds an open borrow into the arena:
```rust
  cs.with_cs_name(|cs_name : &str| {   
    // code using cs_name 
  })
```

But ensuring that guarantee requires an *active* borrow from the threaded arena (behind one lock or another, such as `RefCell` or `RwLock`), so whenever we need more than *one* string value at a time, we get into a risk of dead-locking.

I think I am definitely in "more trouble than it's worth" territory.
[...]

PR #141, opened 2023-04-16
Threaded string interner

dginev:
The theme of this PR is
"The end does not always justify the means"

As I lamented in #140 , Tom made a correct observation that having a write Mutex lock on a global arena singleton will be too slow in practice - as the tests ended up demonstrating after our call ended (2.5 single threaded, up to 10 seconds with full threading due to Mutex lock starvation). This PR maintains the 2.5s single threaded but only need 1.4s to complete in a multi-threaded cargo test (as there are no cross-thread interactions).

Out of sheer stubbornness, I completed a ghastly refactor that branches off my in-progress alignment work and adds a `thread_local! { ... }` wrapper around the string interner, ensuring that each thread will have its own dedicated memory arena.

In practice, that had one major serious consequence:
 - it is no longer possible to take a `&'static str` from arena::resolve, which is generally very easy and simple to use in Rust. Instead, we need a `RefCell<StringInterner<...>>` wrapper, as well as a `.with` method guard which ensures we are thread local. That leads to a very unpleasant reality where we can resolve a limited lifetime `&str`, which must not outlive the ownership of the RefCell lock. Said otherwise, this setup requires passing *in* the closure which will use the borrowed string for the arena to execute, and the `&str` is never released in the outer (latexml) scope.
 -  An example of this awkwardness is:
    ```rust
    let internalcs_str = newcs.with_cs_name(|csname| format!("\\@chardef@{}", csname));
    ```
 - But the worst part is having to actively think about the borrowing scopes. If one casually borrows the string text of a token from the arena, and right afterwards pins a new string back in the arena, there will be a runtime panic due to read+write locks co-existing:
 - ```rust
   let new_token = token.with_str(|text| T_CS!("\\new{text}")); // this will panic
   let new_token = T_CS!(token.with_str(|text| format!("\\new{text}"))); // this will work smoothly
   ```


Using `RUST_BACKTRACE=1` it is relatively easy to spot where the lock conflict happened, and it is indeed possible to resolve them (as mentioned, tests pass here). However, the pain of encountering them gradually leads to using `SymbolU32` tickets more and more often in method signatures, and using actual strings (both String and `&str`) as infrequently as possible. Maybe that's a feature, since that would lead to faster performance, but it also leads to highly "unusual" and "boilerplate-heavy" code.

The contents here are about 2 days of heavy typing, since these refactors end up touching quite a lot of places. It is probably possible to simplify more, reorient more methods towards `SymbolU32` and hide some boilerplate behind macros.

I admit being tempted by the idea of a massively parallel threaded web application, where each latexml process runs with its own thread-local memory arena... But again, does the cost in programming simplicity justify the outcome?

Unclear.


dginev:
The last commit with passing tests before any of the string interning (and the compile-time glob walker) was on April 4, 2023: 9b8a69c8e5e31a839a227ee917a2e5850bb9f745

That managed strings in the good ol' Rust way (with some `Cow`s), following rather directly the good ol' Perl way.

Looking at both the single-threaded `cargo test`, and the multi-threaded tests, both are at 2.0 seconds. Contrast that to 2.4s single-threaded in this PR - it looks like all the indirection may have *added* some overhead.

So, knowing what I know today, I think my next step would be to remove the arena approach entirely, and get back to an ergonomic setup with equally good performance. Maybe the string allocations are too wasteful compared to tralics, but our programming model does much more string juggling than tralics/TeX do. So maybe this is a small price of admission.

dginev:
The timings with this PR are:
```
...
real	0m1.351s
user	0m2.378s
sys	0m1.747s

$ time cargo test --release --tests -- --test-threads=1
...
real	0m2.325s
user	0m2.094s
sys	0m0.222s

$ time ./target/release/rtx rtx/tests/structure/itemize.tex
...
real	0m0.071s
user	0m0.067s
sys	0m0.004s
```

which are specifically comparable to the times in the description of #142 . So the threaded arena yields (roughly) 18% speedup in single-thread testing, 13% speedup in 32-thread testing, and 12% speedup on the single cherry-picked `structure/itemize` test.

It is pretty clear there is a 10% (conservative estimate) speedup to gain by doing the arena approach. The open question remains: Is that worth the trouble?

dginev:
Interesting... more than the performance, the discipline of using `thread_local!` lead to a lot of execution clarity. It is now *accurate* to think of each `core conversion` job as a single-threaded compute task, as far as the backbone infrastructure is concerned. And more than accurate, it is *ensured* by Rust's compile checks over the `Rc<>` guards (which are not shareable between threads). Similarly for the locks, by switching back to the simpler (= cheaper at runtime) `RefCell` locks, the compiler will prevent any cross-thread attempts at locking the same object.

This makes parallel execution engines, such as web servers (as in https://latexml.mathweb.org) or task runners (such as [latexml-runner](https://github.com/dginev/latexml-runner) ) very easy to realize. The engine will be responsible for spawning threads for each request, as usual, and then have the guarantee the conversion will execute in 1 thread. This also follows the TeX expansion model, which is single-threaded.

Post-processing is more interesting, as one can imagine scenarios where multiple threads may want to work over the same XML document - and where we define the programming model ourselves. That is still not implemented here, and can be multi-threaded or single-threaded. Since it is a separate phase from Core, they are independent considerations.

P.S. one can still spawn new threads from e.g. bindings, but they won't be able to share `State` or similar data directly. They can share a *copy* of any relevant data of course.

dginev:
Meeting summary: We may be leaning to a merge here in favor of the "threaded arena" approach. But it would help to have a little bit more documentation of the high level picture first.

dginev:
That said I do need to finish the alignment work, so temporarily I can reunify the branches...

PR #144, opened 2023-05-01
Alignment and friends

dginev:
I am opening a work-in-progress PR as I have reached a stage of porting Alignment.pm and friends where some basic markup is starting to appear in the XML.

I am currently working on completing the "guessing" heuristics and the "normalization" stage over the finished tables. Sizing is mostly deferred until a later time, as there is quite a lot of code to port, while the sizing machinery is still a piece I am actively learning about.

This PR passes `t/structure/article.tex`


dginev:
Incredibly, I think this PR is now ready to merge with a first functioning approximation of Alignment. It newly passes the `t/structure/article.tex` test.

The changeset is clearly massive, and contains some unrelated changes:
 - a string interner (used in hot code paths, not yet everywhere)
 - tweaks to the math grammar and `XM::Ref` to honor trailing puncts
 - improvements to the Digested `is_empty` checks
 - improvements to the Digested sizing logic
 - added interior mutability for the Tbox and List variants of Digested items.
 - basics of handling markers
 - more careful work with what were originally Perl `local` variables with dynamic scope (still more cleanup can be done there).

The alignments follow the Perl implementation, but explicitly name and type each component being passed. Some extra juggling was needed to appease the mutability checks of Rust, and I suspect we have a couple of interesting code paths we can make faster (especially when we keep re-indexing into the underlying array data, instead of using iterators). But all of that can wait till later.

Notably, I haven't added support for colspan/rowspan and that should be tested well before we claim alignments are "done". But they ought to be done for simpler tables.

dginev:
Alright, CI is passing, let's take a step forward 🚀

PR #153, opened 2023-06-03
A medley of add-ons needed for a first arXiv conversion

dginev:
Trying to convert the showcase article for ar5iv at [1910.06709](https://ar5iv.labs.arxiv.org/html/1910.06709)

This PR ports over:
 - `\left`, `\right`, `\frac`
 - fullpage.sty
 - support for `Document::float_to_element`
 - `^` floating marker in binding replacement strings
 - support for `\boldsymbol`
 - Some more infrastructure is needed for `DefParameterType` to execute `before_digest`, `after_digest`. 

TODO:
- support for `{align*}`
- support for `\eqref`

Maybe I can quickly add the remaining bits and have a first Rust-side showcase article...

### DIFF
--- a/rtx_codegen/src/constructable.rs
+++ b/rtx_codegen/src/constructable.rs
@@ -134,7 +134,7 @@ fn compile_replacement_tokens(mut replacement: String) -> Vec<proc_macro2::Token
   let mut floats: String = String::new();
   let mut has_floats: bool = false;
   let float_res = FLOAT_RE.replace(&replacement, |refs: &Captures| -> String {
-    floats = refs.get(1).map_or("", |m| m.as_str()).to_string();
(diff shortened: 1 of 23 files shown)


dginev:
Ok, pretty happy with the upgrades I just added to `DefParameterType`, allowing the same `sub[] {}` ergonomics as `DefPrimitive` and friends have been enjoying for some time.

As a showcase, here is the fully functioning Rust for the `InScriptStyle` parameter (with VSCode highlighting):



It's starting to get rather neat. And maybe a good time to advertise that hovering over any of the variable names in a VSCode instance with the `rust-analyzer` plugin enabled shows the exact type they have. 

The same is true when hovering over the function calls:
image

which also includes a preview of their documentation, if available.

dginev:
Merging and leaving `{align*}` for future days, when `amsmath.sty` becomes a focus.

Issue #143, opened 2023-04-22
Switch to rust stable, when #[thread_local] is stabilized


dginev:
I was doing some more benchmarking and optimizations recently, and I am happy to say the overhead of "thread checking" static variables is now visibly a slowdown factor: ~10% of runtime for a minimal test such as `structure/itemize.tex`

This is due to the way `thread_local!{}` is implemented in Rust, as it can not fulfill enough guarantees at compile-time, and executes some of them at runtime, each time a thread-local variable is accessed.

There is a new upgrade to this feature, realized via a `#[thread_local]` attribute which is pending stabilization in Rust issue #29594. Once stabilized, we can migrate to using it, and ideally shed this overhead entirely. Opening this issue as a reminder on our end.

dginev:
Fun perk: if the `#[thread_local]` annotation really avoids runtime overhead in full, we can actually succeed in emulating the `$STATE` behavior from Perl.

Which in turn may mean avoiding the explicit `state` argument in all kinds of places that may annoy people fond of the Perl brevity...

We could give this an early pilot test by switching to a nightly Rust and adding the annotations, then measuring for performance degradation.

dginev:
With #154 merged, the changes proposed here are now active. I will keep this issue open as a reminder that we need this feature stabilized in Rust in order to switch back to the stable Rust compiler channel.

PR #154, opened 2023-06-04
[nightly] refactor to lightweight #[thread_local] annotations

dginev:
This PR enacts the suggestion in #143 , using the `cargo +nightly` release channel. It is not usable on `stable` Rust, until `#[thread_local]` annotations are stabilized.

I am indeed seeing the expected speedups on tests an examples, varying anywhere between 10%-60% in runtime, depending on example. For the arXiv example mentioned in #153 , this PR lowers the runtime from 250 ms to 150 ms, a whopping 60% speedup.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -15,7 +15,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: stable
+          toolchain: nightly
           override: true
       - name: run tests
         uses: actions-rs/cargo@v1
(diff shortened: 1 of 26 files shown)


dginev:
The CI runs stable rust, hence it will fail here for now.

dginev:
This branch already has conflicts - so it is risky to keep it on the sidelines for too long.

We discussed that the performance upgrades are worth the change, especially when combined with the potential to do away with the local `state`, `stomach`, `gullet` parameters in functions that do not directly use them. 

So I will likely refresh this PR and merge next week, switching the process to `nightly` Rust for the foreseeable future.

dginev:
Alright, rebased on the lastest master and resolved the conflicts.

As mentioned, this switches rtx back to the nightly channel of Rust for the foreseeable future. Let's see if we can next reduce the `state` argument boilerplate...

PR #155, opened 2023-06-09
TeX.pool coverage 

dginev:
This PR is revisiting and extending the port of TeX.pool (accuracy, coverage). It likely won't *finish* the port, but it may get a little closer to a raw load of `latex.ltx`.

The Rust top-level equivalent file is at `rtx_package::package::tex`


dginev:
Alright, my review of the TeX.pool port is completed. The Rust rewrite of it is still incomplete, but at this point more has been ported than remains to be ported.

It "feels" like ~80% is already ported but that is not a real measurement, more of a guess.

The big win of the PR is that I have updated the line numbers in `rtx_package::package::tex` to point to the latest Perl code of `TeX.pool.ltxml`, and have made sure each definition is in the appropriate "inner pool". 

The originally single file ( ~7700 lines) is chunked in Rust in 37 separate submodules.

PR #156, opened 2023-06-12
allow raising a Fatal on 100 Errors; thread-scoped LogState

dginev:
Greetings from the code mines. This PR achieves something very simple:

1. There is now a new struct named `common::error::LogState` and a thread-scoped singleton `common::error::REPORT` of that type. This allows to do the status bookkeeping from `note_status` and friends from within the `common::error` module, and separately from the main State.
2. I have finally ported the behavior where 100 Errors cause a Fatal, and terminate the conversion. But since we model fatal errors through Rust's `Result` types, that means any function that may throw an `Error!` now needs to be able to return a Result, in the case where that error was the hundredth one.

And yet, there is a benefit to using Result types, as we can think about Fatals as first-class data items, and allow some flexibility. Two examples:

First, recognize the 100-error Fatal, but decide to disregard it if it happens in an inconvenient place (e.g. in a `From`/`Into` cast, which aren't well-suited for Result handling). Here's an example of casting to MuGlue from a Tokens, which I currently don't recognize:
```rust
RegisterValue::Tokens(other) => {
  let message = 
    s!("Token register can not be cast into a Glue: {:?}", other);
  // this executes the error code, 
  // but silently disregards any raised Fatal via the .ok() call
  let err = || {Error!("expected", "dimension", None, message); Ok(())};
  err().ok(); 
  MuGlue::new(0)
},
```

Another possible future use is that the Result type allows catching fatals from intermediate callers and enhancing them with more information. E.g. we could catch a Fatal raised from a binding during an absorb call and only add the locator information once it swims up to the Stomach. But that's just hand-waving, I am not planning on redesigning the error reporting setup just yet.

### DIFF
--- a/rtx/bin/rtx.rs
+++ b/rtx/bin/rtx.rs
@@ -13,13 +13,14 @@ use std::result::Result;
 
 fn main() -> Result<(), Box<dyn Error>> {
   if rtx_core::util::logger::init(log::LevelFilter::Info).is_err() {
+    let err = || {
     Error!(
       "rtx",
       "logger",
       None,
-      None,
       "Failed to load logger. Please check rtx_core::util::logger installed correctly."
-    );
+    ); Ok(()) };
+    err().ok();
   }
   let mut argv = env::args();
   argv.next();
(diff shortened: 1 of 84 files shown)


dginev:
I somehow got side-tracked from stating that even though 84 files are changed in this PR, most of the changes are really trivial tedium related to `Result` types, and handling them (the postfix `?` operator).

The shift to the thread-local LogState report also allowed simplifying a first handful of cases that no longer need the `state` argument passed in. My next PR will likely also start using a thread-local `STATE` singleton, which should drastically reduce the number of `state` arguments - and get a little closer to the brevity of the original Perl code.

PR #158, opened 2023-06-22
Thread local maximalism (state, stomach, gullet, model, error::report)

dginev:
Still developing here (and some pieces will need to move before finished).

The PR experiments with using thread-local globals for all singleton data structures in latexml - notably state, stomach, gullet and model. It is likely that `document` will also follow along before the work here is finished.

The branch is still in a workbench state, but some of the first tests are now passing. I'll add more details as things solidify.

In brief, a lot of the signatures are now lighter - the singletons are never passed as parameters, but instead they are directly addressed from the thread-local storage.

I am starting to get some early hints about performance, but so far only the smallest tests pass, so it is not too indicative. But the hello world test runs just a smidge faster (5% ?), which is encouraging.


dginev:
Tests now PASS, with the gullet, stomach, model and state singletons fully (and safely) encapsulated.

A test run of the PR on my machine takes 1.27 seconds in threaded mode and 2.4 seconds on a single thread.

```
...

real	0m1.277s
user	0m2.441s
sys	0m1.517s

$ time cargo test --release --tests -- --test-threads=1
...

real	0m2.406s
user	0m1.939s
sys	0m0.460s
```

Comparing that to an equally fresh run on the master branch, we seem to be at the same performance levels - I see 1.3 seconds in threaded mode and 2.45 seconds in single threaded mode. Maybe there is even a marginal speedup, but not really noticeable on its own.

In summary, the thread-local singleton approach looks viable, as the PR interface is now both safe and without any performance penalties. We have gained freedom from having to explicitly pass around the singletons as arguments, which simplifies pretty much all function signatures and calls - as the PR diff illustrates.

dginev:
Here is one randomly sampled illustration of how bindings look with this change, taken from hyperref:
 - now using ordinary procedural functions instead of object methods
 - simplified signature (though I am still evolving the convention - maybe the inner parens can finally be dropped as well, now that we only expect TeX arguments in there)

image

dginev:
Ready for review.

dginev:
I think we were all cautiously excited to try this direction of data organization in our meeting today, so let's merge and see what we can do with this setup...

PR #162, opened 2024-04-04
compile-time TeX macro

### LINKED ISSUE #111: RawTeX compile-time help
While executing the `RawTeX` directives mostly *has to* happen at runtime, (e.g. some directives are just changing values, as in `\count10=22`), until we adopt the more principled "dump" approach that Bruce is working on in the mainline LaTeXML, there is a simple win to gain.

Namely, switching from the string form to the expanded+compiled Tokens form of the instructions. We already have the `tokenizeable` helper in the `codegen` submodule, so using that for RawTeX should be pretty direct.

Jotting the thought down, until I get back to a clean branch to try it out.

dginev:
fixes #111 , insofar as the tokenization step of RawTeX now happens during compilation. 

For now I kept the old runtime macro and introduced a shorter `TeX!` macro, so that one can quickly compare the behaviors in tricky cases.

Does this have any visible impact on performance? Let's see... not really!

The times of cargo test are quite comparable. It's naturally less compute work, but maybe the work was fast enough that a small test won't really showcase it sufficiently.

But there is quite a lot more of RawTeX to come as bindings get ported over so, it may be worth merging?

### DIFF
--- a/rtx_package/src/package/amsgen_sty.rs
+++ b/rtx_package/src/package/amsgen_sty.rs
@@ -7,7 +7,7 @@ LoadDefinitions!({
   Let!("\\@nx", "\\noexpand");
   DefRegister!("\\@emptytoks" => Tokens!());
   DefMacro!("\\@ifempty {}", r"\@xifempty#1@@..\@nil");
-  RawTeX!(r"
+  TeX!(r"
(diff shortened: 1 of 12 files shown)


dginev:
Let's just go for it for now :>

PR #170, opened 2024-05-29
TeX pool reorganization

dginev:
This PR tracks https://github.com/brucemiller/LaTeXML/pull/2358

Took quite some elbow grease to get everything in place - but the Rust port is now aligned with the TeX pool reorganization in Perl.

I still have a regression or two to resolve, as there were partial updates that get carried through. For example updating the `aftergroup` test necessitates getting the latest `insertBlock` code implemented. Will remove the WIP and merge once all tests pass again.


dginev:
All tests finally pass ✅ 

---

There is some ongoing messiness with the string interning. Using the `SymStr`-typed tickets for string operations forces development into a context switch where I start thinking about managing the string arena workflow, avoiding allocations and worst of all - possible mutability panics at runtime.

I left a couple of comments at the worst offending lines, but this may eventually necessitate a new interface, similarly to how I introduced `SymHashMap<T>` as a convenience abstraction over `HashMap<SymStr, T>`

PR #175, opened 2024-10-07
Expand refactor

dginev:
Tracks latexml [PR 2421](https://github.com/brucemiller/LaTeXML/pull/2421).

@brucemiller you're welcome to commiserate how the 10 changed files in Perl swell up to ~30 files in Rust, as in Rust we don't get the convenience of variable argument list sizes. So each existing caller of `read_arg` and `read_balanced` needed a tweak to the new argument list.

This could have been avoided by introducing new function names (that alias the arguments), or creating a macro dialect -- as macros allow variable argument lists. But for now I'm tracking the original as closely as I can, a cleanup can always follow later.


rtx_core/src/gullet.rs
```diff
@@ -523,22 +525,37 @@ pub fn read_x_non_space() -> Result<Option<Token>> {
...
+#[derive(Copy, Debug, Clone, PartialEq, Default)]
+pub enum ExpansionLevel {
+  // No expansion, reads currently present tokens
+  #[default]
+  Off,
+  /// Expands while reading, but deferring `\the` and `\protected`
+  Partial,
+  /// Expands completely while reading
+  Full
+}
```
dginev, line 538: I also added an explicit `enum` to designate the expansion level, seems to add a bit of clarity over a free-range integer. But it's not set in stone...

Issue #183, opened 2024-10-10
String literals and max-width


dginev:
I am doing some cleanup with rustfmt today, via:
```
cargo +nightly fmt --all
```

Which encounters various examples that are hard to fit in a 100 character line, most notably raw TeX literals from bindings.

A worst case of how verbose things can get is:
```rust
  DefMacro!(
    "\\lx@@typerefnum@@{}",
    r"\@ifundefined{#1typerefname}{\@ifundefined{#1name}{}{\lx@refnum@compose{\csname #1name\endcsname}{\lx@the@@{#1}}}}{\lx@refnum@compose{\csname #1typerefname\endcsname}{\lx@the@@{#1}}}"
  );
```

My macro DSL is missing a technique to allow chunking such constructs into multiple lines, elegantly and without bad breakage. The key missing feature is to trim away any leading whitespace, as TeX can be very sensitive to introducing extra tokens.

The native [string literals in Rust](https://doc.rust-lang.org/rust-by-example/std/str.html#literals-and-escapes) have some unfortunate trade-offs for latexml.

But there is a crate called [indoc](https://docs.rs/indoc/latest/indoc/#using-indoc) which implements some of the conveniences we are after. It could be a good inspiration to improve the macro mini language.

dginev:
This is a less-ambitious framing of the DSL issue in #171 , and a bit more TeX-oriented in motivation. But naturally the XML replacements also suffer from the 100 char line limit.

dginev:
For now I am setting `error_on_unformatted = false` in `rustfmt.toml` to suppress these cases from getting checked, but this will hurt readability (as seen in my example above).

dginev:
We just "deal with it" for now, not that big of a problem.

PR #184, opened 2024-10-10
Reformat entire codebase

dginev:
A good next step after completing a rename -- reformatting with Rust's prettifier.

I opened a related issue at #183 for the cases that don't have an obvious chunking strategy to <100 chars per line.


dginev:
I studied the latest-and-greatest [rustfmt Configurations](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md), and updated to the 2024 Style Guide, which appears a lot more compact.

Added some other space tweaks that I enjoyed, until someone with different taste comes along...

PR #200, opened 2026-03-08
Pass 3 theorem tests using Opus 4.6 

dginev:
This PR passes 3 new subtests from 55_theorem:

55_theorem
- [x] ├── amstheorem.tex
- [x] ├── latextheorem.tex
- [ ] ├── ntheorem.tex
- [x] └── theorem.tex

---

@nicodjimenez was right after all - agents can now meaningfully drive the Rust port of LaTeXML - with occasional human supervision.

This PR was prepared via a command-line `claude` session with Opus 4.6, **high effort** setting. The results so far took in the range of 100,000-150,000 tokens, and about 4 hours of semi-supervised progress.

I am quite impressed that the model can make incremental progress by successfully tracing breadcrumbs from very distant parts of the code base, tie together the differences between the original Perl and the current Rust of each step, while also satisfying the Rust compiler. It is still significantly slower than me in the exploration phase -- as it doesn't understand LaTeXML's design in full. But it is decidedly **faster** than me in the coding phase, when it prepares the Rust changesets.

I think I will postpone the work on ntheorem.tex, as there is a lot of code drift between the state of the 2024 Rust and the 2026 Perl -- an update and refactoring of the currently passing tests could be a better use of time.

But this method appears to now work, and if we find a Claude Max grant it may be possible to complete the port as a background task to my main work.


dginev:
It is also good to call out how difficult our tests are for incremental TDD -- the theorem tests include a lot of unrelated internals (such as math parsing for `\partial`) and getting them to pass required 10+ independent unit-level changes by Claude. In very different places. I was impressed that I could do it when I started, doubly impressed that Claude can.

PR #203, opened 2026-03-15
Pass >200 tests (61%)

dginev:
### Summary

- Port IEEEtran.cls binding from Perl (382 lines): conditionals, keywords/proof/biography environments, IEEEeqnarray redirect, L/C/R
    column types, list overrides, QED support
- Port colortbl.sty binding: colored table rows/columns/cells with backgroundcolor attribute
- Port amstext.sty binding, fix \color tex attribute reversion, un-ignore mathcolor_test
- Port myxkeyval.sty binding, un-ignore xkeyvalstyle_test
- Implement hyperref RDF metadata, fix xcolor conditionals, add neutralizeFont to tags
- Add new DefMacro!(T_CS!(), literal_params, literal_expansion) pattern to avoid infinite loops with starred commands
- Fix DefPrimitive None replacement to emit isEmpty Box with CS reversion (essential for \rm, \it font switches in tex attributes)
- Add 18 new test suites from Perl: pgf (2), tikz (10), moderncv (2), expl3 (2), slides (2)
- Comprehensive diff scan of all 65 ignored tests; revised SYNC_STATUS.md work plan by feasibility tiers
- Update README: 77% ported tests, alpha stage

### Test plan

- Full test suite: 214 pass, 0 fail, 65 ignored
- No regressions from isEmpty Box fix or IEEEtran binding
- All new tests properly ignored with descriptive comments

🤖 Generated with Claude Code

### Disclosure

Used about 48 hours of Opus 4.6 "heavy effort".

The autopilot is getting a little less stable now, as some of the more advanced failures can cause infinite loops in testing and even infinite loops in compilation. An example of that was `DefMacro!("\foo*" ...` where the `*` character fooled the macro machinery to keep expanding. I think. I still need to troubleshoot and fix that fully, it likely requires human intervention.


dginev:
Follow-up: The automated count was overinflated at 77% test coverage due to Opus only counting `40_math` and `70_parse` as individual tests. They are top-level groups with many tests underneath, as can be seen in https://github.com/dginev/latexml-oxide/issues/30#issuecomment-467065264

PR #208, opened 2026-03-25
Pass 300 tests (92.6%) - full math parse suite, 499 package bindings

dginev:
### Summary (generated)

This round brings latexml-oxide to 300/324 tests passing (92.6%). The entire 28-test math parse suite now passes. Package coverage reaches 499 bindings (408 core + 91 contrib).

Highlights

- 300 tests pass, 0 fail, 24 ignored (was ~285 pass)
- 28/28 math parse tests pass — the full 70_parse suite
- 499 package bindings — 408 in latexml_package + 91 ar5iv contrib bindings
- 338 new files, 417 files changed, +11,431 / -2,431 lines

### Key breakthroughs

Fenced XMDual nesting fix — The biggest single improvement. Comma-separated lists inside parentheses like `(1+, 0+, 1-, 0-)` now produce per-item XMRefs in the XMDual structure, matching Perl's NewFenced
exactly. This alone reduced the calculus test from 332 diffs to 0.

OOM root cause found and fixed — parse_parameters entered an infinite loop on non-word characters in CS names (e.g. `\nprt@sign@+`). The regex `PARAMSPECT_CHECK_RE` matched empty strings on +/-, never advancing.
Fixed with Perl's literal Token fallback + 50-step safety guard.

Grammar: 8 new math patterns
- Dirac bra-ket notation: `|ψ⟩` → `ket@(ψ)`, `⟨φ|` → `bra@(φ)` with dedicated qm_ket/qm_bra semantic actions
- Fenced singleton operators: (`∫`) → integral, (`Δ`) → Delta
- Scripted OPFUNCTION/TRIGFUNCTION absorption: `\log_e a` → `(log_e)@(a)`
- Compound operator pruning: `\nabla\log x` → `(nabla@(log))@(x)`
- Operator as standalone term: `D - 1` = subtraction
- XMArg lexer unwrapping: `a_{ij} → a_(i*j)` matching Perl
- Conditional meaning for (item|item) fence pattern
- Formula list in fenced expressions

Semantic pruning improvements — `apply_invisible_times` now checks scripted Apply bases and compound operator results, correctly preferring `prefix_apply` over multiplication for function/operator expressions.

### Package porting

- 8 previously-blocked modules re-enabled (jheppub, neurips, numprint, revtex4_1, revtex, titling, vmargin, algorithmic)
- 17 new package modules including llncs.cls, pgf/tikz/xy stubs, amstex, rotate, svg, aas_macros, a0poster, aipproc, amsrefs, mnras, revtex3_support, thmtools, svmult
- 91 ar5iv contrib bindings (91% of ar5iv ecosystem covered)

### Documented blockers (for future work)

- C2: `\scriptstyle` fontsize 50% vs 70% — absolute vs relative mathstyle reset
- C5: `\times` vs `invisible-times` precedence in mixed chains
- C8: `|` inside `()` causes exponential Earley ambiguity with ket rules — needs QM subject-area pragma
- C9: MIDDLE fence rules ready but diverge from Perl (improvement needs approval)

🤖 Generated with Claude Code


dginev:
An important note here: A wide swath of tests are intentionally changed as the id numbering has to get re-assigned after a parse with Marpa. Otherwise the trees read very confusingly, as there is a lot of intermediate parse pruning which removes nodes with ids generated on the fly, leaving various gaps.

This decision can be changed at any point later of course, but for now an in-order tree traversal id scheme seems simple and maintainable.

PR #220, opened 2026-04-30
[parity] 10,000 arXiv articles 

dginev:
### Summary (Deyan)

We now match LaTeXML success rates on the "10,000 shuffled articles" sandbox.

### Description (Deyan)

This PR went overboard in time and size, as getting to full coverage had many separate long tail inaccuracies that required various levels of troubleshooting by claude (with intermediate commits). Having arXiv as a test set is a tremendous aid to getting to perfect translation I think, especially when the underlying trickiness comes from subtle language differences (e.g. rounding accuracy, truthiness of non-boolean values, filesystem I/O ...).

The main highlights I would offer are:
 - an extraction and refactor of `latexml_engine` that now tightly follows the latest file organization of LaTeXML in perl
 - much improved reading and writing of the precompiled TeX and LaTeX kernels (aka "dumps").
   - We are closer to a distribution strategy, but not quiet yet. For now I am happy we have them efficiently running as part of CI.
 - growing and making robust the bindings in `latexml_package` and `latexml_contrib`
   - for example, the experimental biblatex binding got fleshed out here.
 - a few meaningful performance optimizations both in core and post-processing.
 - extracting various tricky edge cases in arXiv articles into new unit tests
 - finally reran `cargo fmt`, adding to the diff. Maybe excusable since it was already unmanageable. 
  
The PR is more for historical archival than review at this size, and I am curious to see if squashing will work without me having to rewrite the history.

### Next step

parity with 100,000 arXiv articles.


dginev:
CI is ✅ on what are now 1110 tests. Merging.

PR #231, opened 2026-05-12
[parity] 500,000 arXiv articles

dginev:
### Highlights (Deyan)

We have a clean conversion of the "no problem" subset of arXiv, the "easiest" articles for LaTeXML.

This PR is largely one agentic loop doing conversions over the article set (downloaded on a local SSD drive)
 - resolved many subtle inaccuracies in translation
 - handful of very successful performance-oriented refinement passes
 - improvements to precompiled kernel "make_formats.sh" steps accuracy
 - improvements to graphics post-processing (some forward thinking upgrades by claude in handling epsf, and using mutool for efficient PDF processing)
 - pushed more of the "convenience" bindings claude had put in place to the full "raw interperation" of the original package in texlive.
 
Plenty of commit noise, will squash.


dginev:
And yes, they're all converting to healthy HTML with no errors emitted now.

PR #240, opened 2026-06-03
Fix/regression reports html feedback

dginev:
This PR comes from a dev demo we did today, and likely needs more review than usual 
(discussing while claude-ing lowers my focus).

### Summary(generated)

1. 05905ad — fix(xy): drop orphaned SVG drawing ops at hopeless insertion points
    - Problem: When raw-xy aborts a diagram (e.g. `\Qcircuit` matrix setup rejected — same rejection in Perl), 
      already-digested drawing whatsits resurface at the `\end{document}` unwind inside a stale `<ltx:text>`.
      Every `svg:path`/`svg:g` insert then fails → 83 malformed-insert errors → Fatal:TooManyErrors → paper ships nothing, while Perl ships a degraded document.
    - **Fix**: svg_empty_element and the svg wrapper constructors (`\xyscale@@`, `\xyRotate@@`, `\lx@xy@svgtext`, `\lx@xy@move@to`, `\lx@xy@svgnested`)
        now skip emission in the "hopeless orphan" state (no float host and `can_contain_node_somehow` is `None`) — unreachable for healthy diagrams.
    - Witness: arXiv:2501.02222 goes from 102 errors + fatal + zero output → 28 errors with 588KB XML shipped (Perl: 22 errors). Healthy `\xymatrix` output
    - Also adds an env-gated `LXML_TRACE_BOUND_MODE` diagnostic in end_mode_opt.

2. deb2b57 — fix(post): give empty core-conversion results a `<document/>` root

    - Problem: Perl (LaTeXML.pm L330-336) post-processes even fatally-aborted conversions by substituting a bare `<document/>` root.
        Rust fed the empty serialization straight to libxml → "Got a Null pointer" parse error, and cortex_worker shipped a 0-byte `.html` plus the spurious error on every Fatal paper.
    - **Fix:** Mirror Perl — substitute `<document/>` for empty/whitespace-only input in `run_post_processing` (latexml_oxide/src/post.rs, +8 lines).
        XSLT output is then byte-identical to xsltproc, so the archive shape matches the Perl toolchain.
    - Witness: 2403.19758 (tikz-cd fatal) — clean empty html, no parse noise; 2501.02222 (with fix 1.) now ships 264KB of real HTML end-to-end.

Issue #47, opened 2019-04-03
[Feature] Accurate latex linting


dginev:
This is a "Future" idea, past the 1.0.0 milestone. I am writing some latex and am realizing no traditional linter can be applied completely safely, as the advanced tex markup  can trip up any casual text-munging processor.

Having a version of rtx with: 
 * near-instant conversion
 * token locators persisted to the html via attributes
 * a connection to a classic linter,

One can lint over the plaintext, map it back to the source, and show the mistake in the editor window. (in cases where the match is identical to the source, and not e.g. the result of some complex invocation, say, `\def\au{au} \au{}tor` which results in the misspelled word `autor` for `author`, in which case the source range would have to be over `\au{}tor`).

Something to motivate working on rtx, as reliable tooling over latex is a cool add-on (and easy win for markdown at the moment).

dginev:
Example of the state of art for VScode:  https://github.com/ta2gch/textlint-plugin-latex2e

PR #243, opened 2026-06-04
Persistent LSP server: latexml_oxide --server for editor/preview integration

dginev:
## Scope (grown since opening — merge-once PR)

This PR now carries four bodies of work on the persistent LSP server, in commit order:

### 1. The server itself (original scope) + rebase onto post-#240/#242 master
A JSON-RPC-over-stdio server (LSP framing) for the source-provenance editor track (`docs/SOURCE_PROVENANCE.md`, issues #47/#92): LSP subset (`initialize`, `didOpen`/`didChange` → `publishDiagnostics`, `shutdown`/`exit`) + custom `latexml/convert` returning `{html, log, diagnostics, sources, status, statusCode, root}`. Architecture: **warm-preamble + fork-body** (preamble digested once; each conversion forks a throwaway child over COW state), single-threaded `poll(2)` loop with preemption, shared time+RSS `Watchdog` (124/137 exit codes) reused by cortex_worker and the new `--max-memory` CLI flag, named in-memory mouths for `--source-map` provenance, Perl `push/popDaemonFrame` ported (with honest divergence notes), `serde_json` for JSON. Rebase verified: 8/8 commits replayed identically; both sides' upgrades audited.

### 2. Critical-review fixes (`74c4a15f84`)
Line-by-line audit findings, all fixed: unanswered missing-params requests (-32602 now); didChange-flurry snowball (same-project preemption + pending-queue coalescing); orphaned `setsid` graphics converters on preemption (`PR_SET_PDEATHSIG` + SIGTERM-grace); UTF-8 percent-decoding; `--max-memory` standalone no-op; commented-out `\begin{document}` splitting the preamble; partial-frame stdin deadlock window (poll-driven framing); child-stdout protocol corruption insurance (`dup2 /dev/null`); fork-safety invariant `debug_assert`ed (thread-count baseline); signal deaths as `128+signo`.

### 3. Module split (`fe0019606d`)
`lsp_server.rs` (1.8k lines) → `lsp_server/{mod, json, diagnostics, protocol, document, server, project, overlay, unix, generic}.rs` — pure motion, public path unchanged, tests beside their subjects.

### 4. Multi-file project-root + overlay model (`1f2c055c66` plan, `aa0d3561b7` implementation)
The review's biggest product gap, closed per `docs/LSP_MULTIFILE_PLAN.md`:
- **Root detection**: client override → `% !TEX root` → self-contained fast path → `find_main_tex` (the `--whatsin=directory` arXiv heuristic) with reference-guarded walk-up (a candidate root must `\input`-reference the buffer — no fragment hijacking). Every trigger converts the ROOT; responses carry `"root"`.
- **Unsaved-buffer overlay**: rides the engine's existing `{file}_contents` filecontents channel — **zero engine diff**, COW-inherited by fork children. Warm-cache validity adds a read-log dep snapshot (source table = read-log), pinning each warm-up-opened source at `Overlay(version)`/`Disk(mtime)`.
- **Per-file diagnostics**: record-format log parser (locators live on tab-indented continuation lines — the old parser never saw them), real squiggle ranges, file attribution, grouped per-uri publishes with stale-clear.

## Verification
- Full suite **1385 / 0 / 0** (+~50 tests across the four stages, including 13 multi-file unit tests).
- **`tools/lsp_smoke.py`** (committed live harness): **15/15** across basic / preempt / multifile — incl. the acceptance scenario: didOpen `sections/ch2.tex` with unsaved text → chapter convert resolves to `main.tex`, preview contains the root's content **plus the unsaved edit** (disk text absent), `\undefinedmacroxyz` attributes to `ch2.tex`.
- **ar5iv-editor** (path-dep consumer): full suite 51/0 + all 6 heavy real-conversion tests + live WebSocket conversion through a booted server — green against this branch.
- Clippy clean outside `latexml_math_parser`.

## Known gaps (recorded in SYNC_STATUS § Persistent server)
Unguarded in-process fallback (native hang wedges the server; mitigation: fork the fallback too); warm-root pool / Windows worker story; push-model protocol unification; `generic.rs` compile-checked only on unix hosts.


dginev:
## Critical review + fixes (`74c4a15f84`, `fe0019606d`)

A line-by-line audit of the server (assumption validation + bug hunt) found the architecture sound — warm-fork + single-threaded poll is the right design — plus 7 real bugs and one inaccurate fidelity claim. All findings are now fixed on this branch.

### Assumptions verdict
- **fork-safety**: valid — verified no thread outlives a parent-side conversion (graphics workers are `thread::scope`-joined, `run_with_timeout` is poll-based, Watchdogs are child-side only). Now *enforced*: a `debug_assert` compares `/proc/self/task` count at fork time against a server-start baseline. (First cut asserted "exactly 1 thread" and panicked immediately — the binary always runs the server on a worker thread while `main` parks in `join()`; the real invariant is *no growth*.)
- **daemon frames "faithful port"**: claim corrected — Perl's `pushDaemonFrame` `daemon_copy`s mutable HASH/ARRAY bindings (State.pm L607-627); the Rust port doesn't, so in-place `with_value_mut` mutations would leak through a reset. Unwired today; comment now says exactly what must be added before a thread-reusing daemon relies on it. `pop_daemon_frame` no longer `unwrap()`s an empty stack.
- **in-process fallback**: documented gap (SYNC_STATUS) — it runs with only the cooperative deadline; a native hang there wedges the server. Mitigation candidate: fork the fallback too.

### Bugs fixed
| Finding | Fix |
|---|---|
| `latexml/convert` with missing params never answered → client hangs | `-32602` error response (both servers) |
| didChange flurry → N queued serial full conversions | same-uri didChange/didOpen now **preempts** + pending-queue **coalescing** (`message_doc_uri`/`superseded_in_pending`) |
| preemption SIGKILL orphans `setsid`'d gs/inkscape grandchildren | `PR_SET_PDEATHSIG` in `run_with_timeout` `pre_exec` (+ `getppid` race guard); preemption is now SIGTERM → 50 ms grace → SIGKILL |
| `%XX` decoded char-per-byte → mojibake paths | byte-level decode, `from_utf8_lossy` |
| `--max-memory` silent no-op outside `--server` | standalone CLI watchdog now `with_limits` |
| commented-out `\begin{document}` split the preamble mid-comment | comment-aware `find_begin_document` (`\%`-escape aware) |
| partial-frame stdin block starved the child pipe (3-way deadlock window) | `wait_for_child` is poll-driven: one `fill()` per readiness, frames drained from the buffer |
| child inherits protocol stdout | `dup2(/dev/null, 1)` insurance |
| signal deaths reported as `-1` | `128+signo` (OS OOM-kill → `137`, same as the Watchdog ceiling) |
| recursive `take_frame` on malformed-header floods | loop |

### Refactor (`fe0019606d`)
`lsp_server.rs` (1.8k lines) split into `lsp_server/{mod,json,diagnostics,protocol,document,server,unix,generic}.rs` — pure code motion, public path unchanged, tests moved next to their subjects.

### Verification
- Full suite **1372 / 0 / 0** (+7 new tests: pipe-driven `FdReader` framing, preempts matrix, UTF-8 decode, comment-aware split, coalescing).
- Extended live smoke: initialize → convert(id2) immediately superseded by convert(id3) → id2 `cancelled` (int statusCode), id3 returns real HTML **for the newer text**, malformed convert answered `-32602`, clean `exit` (rc 0).
- Clippy clean outside `latexml_math_parser`.

Remaining known gaps are recorded in the SYNC_STATUS server section: unguarded in-process fallback, non-recursive dependency scan, first-line column offset, `generic.rs` (non-unix) is compile-checked only on unix hosts.

dginev:
Folded the multi-file project-root + overlay implementation into this PR per merge-once preference (`1f2c055c66` + `aa0d3561b7`, fast-forwarded onto `feature/lsp-server`). Description updated to cover the full four-stage scope. Suite 1385/0/0, live smoke 15/15.

dginev:
**Performance review follow-up (2026-06-05): stale-preamble bug found & fixed — `3fe472ff37`**

A per-edit-location cost review ("what does one keystroke cost, where?") found the multi-file landing's warm-cache dep snapshot was **vacuously empty**: it reused the locator `source_table`, which is populated at *document-construction* time (in the forked body child, AFTER the parent snapshots) and filters out `.sty`/`.cls`. Live-confirmed consequence: an **unsaved edit of a preamble-consumed file** (`\input{defs}` in the preamble, a local `.sty`) never invalidated the warm cache — the preview silently kept stale macro bodies until the file was saved.

Fixes in this commit:
- **True read-log** at the `Mouth::create` seam (`state::opened_sources`, `FxHashSet<SymStr>`, ~15 engine lines, one set-insert per file open; `\openin` included). The snapshot **excludes the root** — pinning the root buffer's version would have re-warmed on every body keystroke.
- **Dir scan narrowed to file-SET equality**: comparing mtimes forced a full preamble re-warm on every save of a same-dir *body* file (`.tex` chapter, `.bib`). Appearance/disappearance (find_file shadowing) is all the scan needs; opened-file staleness is the read-log's job.

Verification: full suite **1386/0/0**, smoke **19/19** (new `staledep` scenario guards the regression), live probes: stale-preamble now invalidates, body-file save converts warm (0.05s vs 0.7s cold), root body-edit cache ratio 0.08.

Newly tracked gaps from the same review (SYNC_STATUS § Persistent server): unpreemptible synchronous warm-up (zygote-chain candidate), graphics conversion output landing in the server's CWD (`destination: None` → `dest_dir="."`), `.bib` reads bypassing the unsaved-buffer overlay (plan v2).

dginev:
Tested and good to go. Deploying to showcase next.

PR #284, opened 2026-07-14
release(windows): fully-static .exe via +crt-static (drop VCRUNTIME140 dep)

dginev:
Makes the Windows release .exe truly self-contained: statically links the CRT (+crt-static) with a matching x64-windows-static vcpkg triplet, so it imports only core OS DLLs and runs on any Windows with no VC++ redistributable. Scoped to the release leg; dev/test stay on -md. Adds a dumpbin no-VCRUNTIME CI guard. Verified locally (dumpbin: no VCRUNTIME/UCRT). See docs Phase 5.1.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -392,12 +392,13 @@ jobs:
           if-no-files-found: error
 
   # Windows (x86_64-pc-windows-msvc) — a SINGLE self-contained `.exe`. Uses the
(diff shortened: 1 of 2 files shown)


dginev:
Superseded by the consolidated Windows-release PR (in-process kpathsea + crt-static folded together, since both touch the same release leg).

PR #285, opened 2026-07-14
Windows release: in-process kpathsea + fully-static (supersedes #284)

dginev:
Makes the Windows release `.exe` **fast** and **fully self-contained**, consolidating two improvements into the one Windows release leg. Supersedes #284 (crt-static), which is folded in here.

## 1. In-process kpathsea (the perf fix)
The RC uses the subprocess-`kpsewhich` backend (`KPATHSEA_NO_LINK`), which pays a fixed **~0.5 s/conversion** building the `ls-R` cache in Rust — the RC's dominant Windows overhead on simple docs (the Linux/macOS release binaries link kpathsea in-process and never pay it). Switch to a **static in-process libkpathsea** via the new `kpathsea-build-from-source` feature (rust-kpathsea#21: fetch + compile, LGPL-clean crate).

Measured (warm best-of-3, vs the RC), local self-contained build:

| doc | RC (subprocess) | new (in-process) | speedup |
|---|---|---|---|
| hello.tex | 1,123 ms | 518 ms | **2.17×** |
| mathtools.tex | 2,149 ms | 1,454 ms | 1.48× |
| si.tex | 2,809 ms | 2,352 ms | 1.19× |

## 2. Fully static (from #284)
`+crt-static` + the `x64-windows-static` vcpkg triplet drop the `VCRUNTIME140.dll` dependency. Combined with the static kpathsea link (no `kpathsealibw64.dll`) and static libxml2/libxslt, the `.exe` imports **only core OS DLLs** — runs on any Windows, any TeX distro, no VC++ redistributable.

## Wiring
- `kpathsea` crate: `build-from-source` passthrough (rust-kpathsea#21).
- `latexml_core` / `latexml_oxide`: `kpathsea-build-from-source` passthrough features (off by default; inert off windows-msvc).
- `make_release.sh`: `RELEASE_EXTRA_FEATURES` hook (empty on Linux/macOS).
- `release.yml` Windows leg: in-process + crt-static + a `dumpbin` guard asserting the `.exe` imports no non-OS DLL.
- Temporary root `[patch]` on kpathsea until rust-kpathsea#21 publishes.

## Verified locally
Self-contained `.exe` (dumpbin: only OS DLLs), `--version` OK, 2.17× on simple docs. Depends on rust-kpathsea#21 (merge + publish, or keep the git patch).

## Notes / follow-ups
- LGPL §6: the static kpathsea link carries the same obligation the Linux/macOS static legs already do — a separate docs pass should add kpathsea to `THIRD-PARTY-NOTICES` for all three platforms.
- Residual ~350 ms Windows cold cost is binary-load/Defender (size); trimming the embedded dump window is the next lever.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -392,12 +392,13 @@ jobs:
           if-no-files-found: error
 
   # Windows (x86_64-pc-windows-msvc) — a SINGLE self-contained `.exe`. Uses the
-  # proven windows-ci-manual-trigger.yml recipe (vcpkg static libxml2/libxslt,
-  # triplet x64-windows-static-md) but builds the publish-grade `maxperf` binary
-  # and packages the bare `.exe` (+ `.sha256`) — no tarball/.deb, the user runs
-  # the `.exe` directly. `kpathsea` is NOT linked on Windows: TeX paths resolve
-  # through the subprocess `kpsewhich` (MiKTeX/TeX Live on PATH). The `.exe` needs
-  # only the ubiquitous MSVC runtime (static-md = static C libs + dynamic CRT).
+  # proven windows-ci-manual-trigger.yml recipe (vcpkg static libxml2/libxslt)
+  # but builds the publish-grade `maxperf` binary and packages the bare `.exe`
+  # (+ `.sha256`) — no tarball/.deb, the user runs the `.exe` directly.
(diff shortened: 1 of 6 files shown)


dginev:
Superseded by #286 (merged to main as 45da5c597c), which bundles the in-process + crt-static work here plus backend auto-selection, MiKTeX installer suppression (kpathsea 0.3.2), search-path parity, and the 0.7.4-rc2 bump.

PR #289, opened 2026-07-15
Attribute the vendored native libs, and ship complete notices with every download

dginev:
Prompted by a copy-edit of the README, which turned up a stale test badge and then a much bigger question: does `THIRD-PARTY-NOTICES` actually capture everything we ship?

It didn't. **libmarpa was attributed nowhere**, despite statically linking MIT *and LGPL* code into every binary we have ever released. Four review rounds later, libmarpa was the *easy* one.

**What this file is:** the disclosure for the **`latexml_oxide` binary we release for download** — the tarballs, the Windows `.zip`, both `.deb`s. That was the original premise and the file now says so.

## The blind spot

`cargo-deny` and `cargo-about` both reason over crate **manifests**. A manifest describes a crate's own Rust code — it cannot describe third-party material the crate *carries*. **No manifest gate can see that class**, and it is not a C problem, which is what took four rounds to learn:

| What ships | Its crate declares | What's actually in there |
|---|---|---|
| **libmarpa 8.6.2** | `MIT OR Apache-2.0` | MIT © Jeffrey Kegler — **plus LGPL-3.0** (`marpa_avl.c`/`marpa_tavl.c`, Ben Pfaff's libavl) and **LGPL-2.1** (`marpa_obs.c`, GNU obstack). Its `COPYING` files are sealed inside a tarball, unscannable. |
| **mimalloc** | `MIT`, `LICENSE.txt` © Octavian Oncescu | MIT © **Microsoft Corporation, Daan Leijen**. cargo-about emits a plausible MIT text naming the **wrong holder**. |
| **libkpathsea** | `MIT OR Apache-2.0`, no LICENSE file | LGPL-2.1, fetched at build time — nothing to scan. |
| **unidecode**'s 1.9 MB table | `BSD-3-Clause` © Amit Chowdhury | data generated from **Sean M. Burke**'s Perl `Text::Unidecode` — Artistic/GPL-1.0+. **Pure Rust**, so no native-code audit sees it either. |
| **Ayu palette** in our CSS | *(booked as "ours/CC0")* | rustdoc's values (© The Rust Project Contributors), itself crediting Dempfi's Ayu. |
| **GNU libiconv** in the Windows `.exe` | *(nothing — a vcpkg default)* | **LGPL-2.1+**, statically linked because `vcpkg install libxml2` takes `default-features = ["iconv","zlib"]`. |

Checked and *not* in this class: `stacker`, `psm` (own shims only).

## The delivery gaps

Attribution is worthless if it doesn't arrive. Three artifacts shipped without it, each assembled by a **different tool** that inherited none of the staging done for the others:

- **F7** — `gen_notices.sh` ran only in the `release` job, so only the x86_64-Linux tarball carried the complete file. macOS/arm64 fell back to §1–4; the Windows `.exe` was bare.
- **F9** — the **`.deb`**: `cargo deb` builds from the *committed* §1–4 file, so `apt install` — the path the README calls the easiest way in — installed notices with no §5/§6/§7.
- **F10** — the **container images** shipped **zero** notices and no LICENSE.

## Round 4 — the gates, and libzmq

Three independent reviewers went at the branch. Every finding below was reproduced against the built artifact before being believed.

**Two gates could not fail.**
- **§5**: the `wc -l < 1000` floor was calibrated before §6 existed. Those license texts are 1341 lines, so a notices file attributing **zero of the 66** shipped crates still measured **1729** and passed — *the fix broke the gate meant to detect the absence of the thing being fixed*. The heading needle can't cover it: `about.hbs` emits it outside `{{#each}}`, proving the template ran, not that it rendered.
- **§6**: the needles were `"GNU LESSER GENERAL PUBLIC LICENSE v2.1"` and friends — strings appearing in **no license text**, matching only the title `gen_notices.sh` echoes. With `-f` instead of `-s`, three 0-byte files shipped §6 as three titles and passed.

Both now grep content (crate count in §5's own range; sentences from the LGPL/GPL bodies, each unique to one file), and both are **proven to fail on the case they missed** and pass the real 4096-line artifact.

**libzmq was wrong in every clause.** It said `cortex_worker` links libzmq *dynamically*, against Ubuntu's `libzmq5`, under *MPL-2.0*. In fact `zmq-sys` has `zeromq-src` as an unconditional build-dep and a 16-line build script with no branch: 277 vendored C++ files, `rustc-link-lib=static=zmq`. The vendored tree is libzmq **4.3.4**, whose own `COPYING`/`COPYING.LESSER` are **GPL-3.0 + LGPL-3.0** — ZeroMQ relicensed to MPL-2.0 only *after* 4.3.x. Confirmed against the artifact: `ldd cortex_worker` shows **no** libzmq.so, yet ZeroMQ's C++ source names are inside the binary. Its **static-linking exception** relieves LGPL-3.0 §4/§5 and GPL-3.0 §6 → attribution, **no relink**; §3.5 now says so rather than leaving it inferred. Not in `latexml_oxide` at all, so it is marked as the one cortex-only exception.

**Why nothing caught it** — the audit pinned itself to the `runtime-bindings` graph and called it "the shipped tree". `zmq-sys` appears **0** times there, **2** in the `cortex` graph the published worker image ships. `gen_notices.sh` already had `NOTICES_CARGO_FEATURES` *because the images link different graphs* — the notices generator knew, the tripwire didn't. It also used `-e normal`, justified by "build-deps are not linked in" — false for the `*-src` shape, and the second half of the miss.

**The container scope was false.** `ldd` on the shipped image shows the binary links libxml2/libxslt/libkpathsea **dynamically** from Ubuntu, plus libgcrypt (LGPL-2.1+), ICU, zlib, liblzma. So *"if you got a binary from us, it has kpathsea's LGPL code in it"* was false for 2 of 7 artifacts — **the sentence fixed earlier in this same PR, wrong the same way on the second attempt**. Containers are now descoped to the standard posture: apt-built aggregates whose libraries carry their own `/usr/share/doc/<pkg>/copyright`, stated once instead of restated per library.

## What changed

- **Attribution** for everything above (§2.3, §3, §3.6), the LGPL relink note + pinned sources (§3.5), the verbatim LGPL-2.1/LGPL-3.0/GPL-3.0 texts (§6, via `licenses/` — byte-identical to Debian's canonical copies), and per-artifact provenance (§7). GPL-3.0 is present **only** as LGPL-3.0's base text; `licenses/README.md` says so, since a folder of GPL files beside a CC0 LICENSE invites the wrong conclusion.
- **Every download carries it**: a `notices` job assembles once for all packaging legs; `make_release.sh` swaps the path `cargo deb` reads and **reads the notices back out of the built `.deb`**; the Dockerfile has its own per-image `notices` stage.
- **Windows ships a `.zip`** instead of a bare `.exe`. The asset URL changes `.exe` → `.zip`; 0.7.4 hasn't shipped, so no published link breaks.
- **`tools/audit_vendored_natives.py`** — a CI tripwire for the class no manifest gate can see. It is a **tripwire prompting a human, not an oracle**: markers not counts, fail toward flagging. Now audits the **union over every published graph** with `-e normal,build`.
- **README copy-edit** (the original ask): tests badge `1533` → **`1577`**, heading levels, `MacOS` → `macOS`, comma splices, license scope.

**Our CC0 is unaffected** — it covers our source, which is exactly what discharges the LGPL relink condition.

## Owner decisions (settled — not open questions)

1. **LGPL posture** (§D.3) — source-availability discharges the static-link relink obligation; no object files, no written offer.
2. **`libmarpa-sys`'s manifest** stays `MIT OR Apache-2.0` upstream — the `-sys` convention. The control is §3.3 + the CI audit.
3. **unidecode** (§D.4) — attribute Burke, record the ambiguity, treat the mapping table as uncopyrightable data.
4. **Windows libiconv** (F13) — keep the vcpkg features and attribute.
5. **Scope** — the file is `latexml_oxide`'s disclosure; containers descope to their in-image copyright files; libzmq keeps a short, correct, clearly-scoped entry because no apt package describes the vendored static 4.3.4.

## Verification

- Completeness sweep: **every** native crate in the downloads graph maps to a disclosure (§3.1–3.4, §3.6), and the cortex-only crates are **absent** from that graph.
- Both new gates proven to fail the case they missed and pass the real artifact (4096 lines, 66 crates).
- Audit reports **18** native crates, identical from the repo root and a subdirectory; each of the 9 newly-surfaced build-tooling crates carries a verdict taken from reading the file.
- Both container images built and their notices read back out of the runtime image; `--build-arg GITSHA` proven required (§7 refuses `commit: unknown`).
- `licenses/*.txt` byte-identical to Debian's canonical FSF texts.

## What this PR does *not* prove

`release.yml` triggers **only on tag push**, so **CI never executes it**. The `notices` job, the `.zip` packaging, the artifact plumbing, and the `.deb` readback are unexercised by the green checks — tagging rc3 (a draft prerelease) is their first real run. `cargo-deb` isn't installed locally, so the `dpkg-deb` readback first executes in CI.

**Three holes shipped knowingly open** (owner decision 2026-07-15: merge as-is, track — recorded in `LICENSE_INVENTORY` "Known holes, shipped open", each with its fix). None affects what the released `latexml_oxide` binary *discloses*; each is a way a future change could go unnoticed:
1. **§7's provenance is not exact by construction** — `marpa` carries no `rev=` and `Cargo.lock` is gitignored, so every runner re-resolves. This PR split `notices` into its own job (on `main` it was a step inside the build job), so §7 records the *notices runner's* marpa commit while five build legs resolve independently. A regression this PR introduced, in the section whose purpose is exactness. Window: hours, on a repo we control.
2. **The `.deb` readback is fail-open in its own scenario** — when `.dist` is missing (the F9 bug itself) the gate doesn't run, it warns. CI-unreachable today.
3. **No gate covers vcpkg feature defaults (F13)** — the audit reasons over the cargo graph, so a vcpkg port default is structurally invisible. `vcpkg list` diffed against an allowlist would close it.

Each of the four rounds found real defects in the previous round's fixes, twice reproducing this PR's own bug class. That rate is worth weighing.


dginev:
## Gaps addressed

**F8 (new): `resources/RelaxNG/svg/` isn't public domain.** §B claimed all 108 RelaxNG files were "PD (NIST)". **83 are the W3C's SVG 1.1 schema** (© 2001–2002 W3C) with Mozilla's modifications (© 2007), and `latexml_core/build.rs` embeds every one unfiltered. The grant is permissive but requires the notice to travel with copies — now attributed in §2.2, §B corrected. Rest of `resources/` swept: nothing else third-party.

**F4: a standing backstop.** `tools/audit_vendored_natives.py`, wired into CI `lint` beside cargo-deny — whose scope limit it exists to cover. Fails when a shipped crate links native code with no §D.2 entry, or an embedded resource carries an unaudited notice. It resolves paths via `cargo metadata`; a find-by-dirname sweep silently misses git-sourced crates like `libmarpa-sys` (my first one did). Both tripwires verified to actually trip — a gate that cannot fail is not a gate. Markers, not counts, for the resource half: a count gate fires on every legitimate new schema file while missing what matters, a new copyright holder.

**F5/F6: §3.5 was overstated.** kpathsea is genuinely pinned (`KPSE_REF`), but the `marpa` git dep carries no `rev =` — the built revision lives only in the gitignored `Cargo.lock`, so a rebuild resolves whatever branch HEAD happens to be. Rather than disturb the local-`[patch]` workflow, new §7 records each artifact's own inputs (oxide commit, marpa commit, `KPSE_REF`). `gen_notices.sh` refuses to write if either fails to resolve, and the release job rejects `unresolved` — an empty relink pointer is a promise we can't keep.

**Verification.** Real `cargo-about` artifact = 3975 lines; the release gate passes verbatim. §7's three revisions cross-checked against lockfile / `build_static_kpathsea.sh` / `git HEAD`. `licenses/*.txt` byte-identical to the FSF texts.

**Still yours:** the F5/F6 posture — whether source-availability satisfies the relink obligation for a static link, vs. prelinkable objects or a written offer. Engineering is done; what remains is a legal call about *how* we discharge the duty, not *whether* we may release.

dginev:
## All three decisions settled — PR is complete

1. **LGPL relink posture → F5/F6 CLOSED.** Recorded in §D.3 as an owner-approved position, mirroring the §C block that settled the TeX-dumps question: source-availability discharges the static-link relink obligation (our source is CC0 and public; §7 records every input commit per artifact). No prelinkable object files, no written offer. The block also names what would force a revisit — an LGPL library whose source *isn't* public and pinned, or our source ceasing to be CC0.
2. **`libmarpa-sys` manifest → left as `MIT OR Apache-2.0`.** Recorded under F6 so it isn't re-litigated: declaring only the wrapper's license is the `-sys` convention, the crate isn't on crates.io so nobody downstream is misled, and an "honest" combined field would trip our own `deny.toml` for no gain. **The control is §3.3 + the CI audit, not the manifest** — which is why the audit exists.
3. **`Cargo.lock` → stays gitignored.** Recorded in §A: §7 meets the licensing need, and committing it is a reproducible-build / supply-chain question deserving its own PR, not a rider on a licensing fix.

**One more hole closed while wiring this up.** `CI.yml` triggers on `push: [main]` + `pull_request` — **so a tag push never ran the vendored-natives audit**, and a release could have shipped a newly vendored native library no gate ever looked at. The `notices` job now runs the audit too, and every release passes through it:

| | main / PRs | tag push (release) |
|---|---|---|
| `cargo-deny` (manifests) | CI `lint` | — |
| `audit_vendored_natives.py` (vendored C + embedded resources) | CI `lint` | `notices` |
| notices content gate (§2.2/§3/§5/§6/§7, no `unresolved`) | — | `notices` |

**State:** `LICENSE_INVENTORY.md` has no open owner-decisions (F1–F8 all closed or landed). Audit green, notices generate at 3975 lines, all gates pass.

PR #300, opened 2026-07-18
fix #299: dump staleness check compares TeXLive years, not kpathsea patch strings

### LINKED ISSUE #299: latexml_oxide: 0.7.4  Where is `tools/make_formats.sh` ?
when compiling with latexml_oxide, I keep seeing this message

```

Warning:latex_dump:mismatch TeXLive MISMATCH — dump stamped `kpathsea version 6.4.1`, ambient kpsewhich `kpathsea version 6.4.2`. The dump may reference macros unknown to the current TeXLive (or vice versa). Run `tools/make_formats.sh` to regenerate against ambient texlive. Silence with LATEXML_SKIP_DUMP_STAMP_CHECK=1.

```
First, I see no tools/make_formats.sh anywhere.  

Can this message be just ignored  using `LATEXML_SKIP_DUMP_STAMP_CHECK=1` ? Or does one need to worry about it?

I am using texlive 2026 on Linux mint 22.3

`

dginev:
**Diagnostic.** The standalone release binary warned `latex_dump:mismatch` on every run: `build.rs`'s generated `compare_stamp_to_ambient` compared the exact `kpathsea version X.Y.Z` stamp string, so a within-year patch bump (release-build container's `6.4.1` vs a user's shipped-TL2026 `6.4.2`) tripped it even though the embedded **TL2026** dump is correct for a **TL2026** install. The remedy it printed, `tools/make_formats.sh`, doesn't exist for a source-tree-less binary user.

**Approach.** Compare TeXLive **years** (already how the dump is selected — `select_embedded` returns a `matched` bool) instead of kpathsea strings, via a testable `dump_paths::dump_year_mismatch_warning(dump_year, ambient_year, from_embedded)`: warn only on a genuine year mismatch (fell back to another year's dump), silent when years match or the ambient year is undetectable. Embedded-path message drops the `tools/` reference; the dev/on-disk path keeps it. The kpathsea string is not a reliable cross-year discriminator anyway (CLAUDE.md: TL2023/TL2025 share it).

**Validation.** Guard `dump_paths::stamp_check_tests` (6 cases incl. the reporter's silent TL2026/TL2026 case); e2e confirmed silent-on-match + still-warns-on-genuine-mismatch. Full suite 1594/0; `clippy --workspace --all-targets -D warnings` + `fmt` clean. Rust-only distribution machinery — no Perl analogue.

Closes #299


xworld21:
Just a note, the LaTeX format does change throughout the year, so checking for TL2026 is not accurate enough. Past years are fine since they are frozen. Running latex on TL2025 shows
```
LaTeX2e <2025-11-01>
L3 programming layer <2026-01-19>
```
or checking `\fmtversion` in TL2026 will bring up 2026-06-01.

dginev:
True, I'm aware.

The only reliable approach to match a moving distribution is to build with the distribution itself. That is possible with an installation from source on a user's machine, but the approach that is friendly to unfamiliar end users would be to bundle the executable with each texlive release itself, or have tighter guarantees for OS package distribution.

I suspect it would now be fruitful to start a discussion with the texlive maintainers about how best to track kernel changes in texlive. I'll talk to @norbusan about this, ideally we have a similar strategy for the arXiv updates in the long-term.

And one more note - the [Dockerfile installation](https://github.com/dginev/latexml-oxide/blob/de433ed877becba101a3c317ad0b63d6d8476bba/Dockerfile#L204-L206) released with the project demonstrates the tight version-matching workflow is possible, just not possible when releasing a ready binary file.

norbusan:
I suppose you mean LaTeX kernel changes. This is not something we as TeX Live maintainers actually control. The LaTeX team makes releases in regular intervals (more or less regular).

But there are more ingredients to a format dump than the latex kernel itself. From the `texlive.tlpdb`, package `latex-bin`, we see:
```
execute AddFormat name=pdflatex engine=pdftex \
      patterns=language.dat \
      options="-translate-file=cp227.tcx *pdflatex.ini"  \
      fmttriggers=l3backend,l3kernel,latex,firstaid,babel,cm,hyphen-base,knuth-lib,latex-fonts,tex-ini-files,unicode-data,dehyph,hyph-utf8
```
which means that on package updates of one of those packages listed, the format is rebuild. That might not change the latex kernel version, but the format.

The format rebuild is then kicked off on the user computer when any of these packages is installed by `tlmgr`.

What would be a useful integration with latexml-oxide and rebuild?

dginev:
> The format rebuild is then kicked off on the user computer when any of these packages is installed by tlmgr.

That is a bit too frequent... But I guess workable. If texlive could also run `latexml_oxide --init` to update its own dumps when this happens, that kind of tight integration could work one day. Some allowances need to be made so that `latexml_oxide` uses texlive-produced dumps instead of its embedded ones, then it should be workable.

xworld21:
> If texlive could also run `latexml_oxide --init` to update its own dumps when this happens, that kind of tight integration could work one day.

I was only suggesting to key the format by at least `\fmtversion` and probably L3 layer, since that seems the most likely origin of syntax(-like) errors. However, there are many other options available, e.g. the dumps could remember location and mtime of latex.fmt and the likes.

I may have overreacted after all: how deeply do packages depend on a specific kernel + L3 layer? Aren't they supposed to check `\fmtversion` anyway to provide some backward (and forward) compatibility? I don't know the answer, but maybe it doesn't really matter that the dumps match the kernel so precisely. Although a message or warning that the format is out of date could be nice (say a naive mtime comparison of latex.fmt and dump? it wouldn't work for Nixpkgs, but that's the outlier, and I can fix that).

PR #333, opened 2026-07-21
fix #321: LookupDefinition + before/after hook pushes for Rhai bindings

### LINKED ISSUE #321: add LookupDefinition to Rhai bindings
In Perl, constructors can be found via `LookupDefinition` and one can attach further behaviours by pushing new closures to the `beforeDigest`, `afterDigest` lists. For instance, BookML patches horizontal and vertical rules to mimic TeX a little better via:
https://github.com/vlmantova/bookml/blob/59a5bd7206e7175d174c05c0b08bbb2451d6a6be/bookml.sty.ltxml#L798-860

Ideally, Rhai bindings should be able to call `LookupDefinition`. If this is too generic for Rhai, some facilities for adding hooks to `beforeDigest`, `afterDigest`, and maybe allowing both prepending and appending, would be very useful.

dginev:
Exposes Perl's `LookupDefinition` to the runtime (Rhai) binding surface, plus the BookML idiom of pushing hooks onto an already-installed definition.

- `LookupDefinition(cs)` returns a `Definition` proxy, or `()` for an undefined control sequence (mirroring Perl `undef`).
- The proxy offers `push`/`unshift` methods (append and prepend) for every before/after hook family — digest hooks on any constructor/primitive/math definition, and construct + body hooks on constructors.
- Pushing a construct/body hook onto a definition kind that never runs it is a clear script error rather than a silent no-op.
- Each push clones the current definition, splices the hook, and re-installs globally — so the patch is active immediately, accumulates across pushes, and persists across groups, faithfully matching Perl's in-place mutation.

Covered by contrib unit tests (undefined, accumulation, prepend order, wrong-kind error) and an end-to-end conversion test proving both construct hooks fire on a live document, in order. Shipped example and docs updated. Completes the Rhai binding-API cluster.


latexml_contrib/src/script_bindings/tests.rs
```diff
@@ -510,6 +510,64 @@ fn command_output_runs_by_default_and_is_blockable_via_env() {
   latexml_core::reset_thread_engine();
 }
 
+/// #321: `LookupDefinition(cs)` returns a proxy onto an installed definition;
+/// `push<Hook>`/`unshift<Hook>` splice a trampolined hook onto its list — the
+/// Rhai analog of BookML's `push(@{ $$def{afterDigest} }, sub{…})`. Perl mutates
+/// the shared blessed def-hash in place; we clone the current front def, splice,
+/// and re-install at GLOBAL scope, so sequential pushes ACCUMULATE and `unshift`
```
xworld21, line 517: @dginev this is possibly wrong. When Perl updates the definition in place, the definition should be updated in the scope where it was made, i.e. at what Knuth calls 'same level', which I was tentatively calling a `scope => 'definition'` assignment last year, when we were figuring out the problem with box registers. That's neither global, nor local. In principle, this could have bad side effects if the definition was genuinely in a local scope.

In practice, at the moment I am the only user of such hooks, and I only use it on definitions that are already global... so the impact is nonexistent.

PS: if you are receiving this twice, that's because I tried commenting on the commit, and from my point of view that comment disapeared!

dginev: Surprisingly this was *my* fault since it appears that claude understood you correctly the first time. I should stop interfering ... Fix incoming, we can definitely do the reassignments in-place.

PR #344, opened 2026-07-22
feat: raw-load Perl-bound style packages (sTeX) instead of misreading the .ltxml

dginev:
**Diagnostic.** Raw-loading a style package that also ships a Perl LaTeXML binding (`<pkg>.sty.ltxml`) misread the Perl source as TeX. sTeX's `stex.sty.ltxml` is the witness: `find_file` preferred the `.ltxml` (kpsewhich lists it) over the raw `stex.sty`, and the loader tokenized Perl as TeX — `$out =~ s/^\s+//;` → "Script ^ can only appear in math mode", `\DefMacroI`/`\stex@backend` undefined.

**Approach.** A `.ltxml` is always Perl and latexml-oxide can never read one; the Rust equivalent of "is there a binding for this name?" is the binding **dispatcher** (compiled + Rhai), already consulted by `find_file`'s `notex` fast-path. So the disk `.ltxml` candidate is dropped — a file we intend to open/read never resolves to a `.ltxml`, and a missing binding falls through to the raw `.sty`, matching pdflatex. With that fixed, raw sTeX 3.x needed standalone.sty's two unconditional requires restored: `xkeyval` (`\define@key`) and `currfile` → `filehook` (the `\AtEndOfPackageFile` hooks) — added unconditionally (a binding emulates its package identically regardless of `INCLUDE_STYLES`). `currfile` moves latexml_contrib → latexml_package (a package binding must not depend on a contrib-only one).

**Validation.** Raw `stex.sty` — and an `smodule` with a `\symdef` symbol — convert with **0 errors**, semantics preserved (`\symdef` operators carry `meaning=`). Guards: `15_stex_raw_ltxml` (self-contained standalone→filehook chain; a stex witness that skips where TeX Live lacks stex). Full suite **1644/0**; clippy + fmt clean; pdflatex parity confirmed.

cc @Jazzpirate — raw sTeX 3.x now loads cleanly under latexml-oxide.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,16 @@
     previously hit OOM/timeout; `scrartcl` `\titlehead` frontmatter capture; and
     frontmatter bindings for `fairmeta`, `selfevolagent` and `openmoss` classes.
   - **A deferred package-load miss no longer poisons a later raw load.**
+  - **Perl `.ltxml` bindings are never read as TeX.** When raw-loading a style
+    package that also ships a Perl LaTeXML binding (e.g. sTeX's `stex.sty.ltxml`),
+    file resolution used to hand back the `.ltxml` — Perl source — and the loader
+    tokenized it as TeX, emitting spurious errors. latexml-oxide can't run Perl
+    bindings; binding availability is decided by its own dispatcher, so a missing
+    binding now falls through to the raw `.sty`, matching pdflatex. The
+    `standalone` binding also regained standalone.sty's real `xkeyval` +
+    `currfile → filehook` dependencies, so the package-file hooks
+    (`\AtEndOfPackageFile`, …) and `\define@key` exist — enough for raw sTeX 3.x
+    to load cleanly.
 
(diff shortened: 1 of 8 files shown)


dginev:
[fyi] this was all Opus catching slip-ups from the large writing sessions, with a couple of minimal test checks from me.

Jazzpirate:
nice :3 (although note that stex is by now on version 4.1, but I expect that should make no difference for latexml).

I would expect latexml on stex to behave like pdflatex - which means it would throw all the semantics away, which is reasonable a priori. In the long term, it would be interesting whether we can build a latexml-specific backend config that preserves the semantics, by implementing a `stex-backend-latexml.cfg` analogously to those for [pdflatex](https://github.com/slatex/sTeX/blob/main/tex/etc/stex-backend-pdflatex.cfg) and [rustex](https://github.com/slatex/sTeX/blob/main/tex/etc/stex-backend-rustex.cfg), respectively... (a few macros "insert raw html node `<text>` here" and "attach attribute A with value V to the node generated by `<text>`")

PR #373, opened 2026-07-25
feat(math): parse fences split by TeX's null delimiter instead of giving up

dginev:
**Diagnostic.** `\left( … \right.` split across an alignment break leaves each cell with one unmatched delimiter. `fenced_array`'s `open_fenced`/`close_fenced` rules look like they cover this, but `array` is `token!(array ~ "ARRAY")` — the XMArray terminal — so they only ever applied to an unmatched fence around a *matrix*; general-expression fences are balanced-only. The grammar returned **zero** derivations and the entire formula, fractions and subscripts included, fell back to `ltx_math_unparsed`.

**Approach.** Re-supply the delimiter TeX says is there and retry, rather than adding open-ended grammar rules that would make every balanced `( … )` ambiguous with a dangling reading. The synthesized `XMTok` has the fence role but empty text, so it renders as nothing and `tex=` reversions stay verbatim. **Gated on total parse failure**, which is what makes it safe — a formula that already parses is unreachable for it. That gate is load-bearing: an unconditional version read the `⟩` of a ket `|f⟩` as a dangling close and prepended a bogus `(`, breaking formulae that had been fine.

**Validation.** Guard `split_fence_test` pins the split fence, a balanced control, and the ket regression. Suite **1684/0** (baseline 1683/0, +1 new); clippy/fmt clean; `si.tex` 4.07s vs 4.02s on main — within run variance.

Measured over 24 arXiv 2606 papers containing the pattern, branch vs main binaries:

| | main | branch |
|---|---|---|
| `unparsed_math` | 177 | **102** (−42 %) |
| errors + fatals | 35 | 35 |
| papers improved | — | 19 / 24 |
| papers regressed | — | 0 |

Witness 2606.13010: 2 unparsed → **0**, with real `<mfrac>`/`<msub>`. The pattern occurs in **4 %** of papers (24 of 600 sampled) — an earlier 9.2 % figure quoted here was wrong, from a looser probe matching `\left.` anywhere rather than a line-ending `\right.`. The remaining 102 are other unparsed causes plus streams too lopsided for the 2-delimiter cap; this fixes a slice, not all of it.

This makes Rust parse input Perl cannot (Perl reports 4 `not_parsed` on the minimal case), so it is documented as divergence #67.

### DIFF
--- a/docs/parity/OXIDIZED_DESIGN_DIVERGENCES.md
+++ b/docs/parity/OXIDIZED_DESIGN_DIVERGENCES.md
@@ -1958,6 +1958,41 @@ branch in Perl, faithfully ported in `document.rs::append_tree`. Perl simply nev
 chains its parser to it. Accepting fragments therefore inserts more of the author's
 content correctly and can never emit *fewer* errors than Perl.
 
+### 67. A fence split by TeX's null delimiter is re-balanced before it is given up on
+
+**Perl.** `\left( a+b \right.` is a fence whose right delimiter is *empty*, and
+digestion emits no token for the `.` at all. When such a fence is split across an
+alignment break — the standard way to break a long parenthesised expression:
+
+```latex
+\begin{align}
+  H & = \frac{\hbar}{2} \left( \Omega_{IX} IX + \ldots \right. \nonumber \\
+    & \quad \left. + \Omega_{ZX} ZX + \ldots \right).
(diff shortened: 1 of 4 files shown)


dginev:
retriggering CI after a repo-wide Actions startup_failure window (~12:27–12:39, hit unrelated branches too)

PR #379, opened 2026-07-25
fix(bibliography): break the self-referential \let that hangs multi-\datalist .bbl files

dginev:
Closes SYNC_STATUS **R4** — the biblatex `.bbl` `TokenLimit` loop (arXiv 2605.17646).

## The bug

`setupPseudoBibitem` unconditionally captures `\save@bibitem` ← `\bibitem` before installing its missing-`\bibitem` redirection. Run it while that redirection is still armed and all three saves capture the *redirectors*: `\save@bibitem` becomes `\restoring@bibitem`, whose body is `\let\bibitem\save@bibitem…\bibitem` — it points `\bibitem` back at itself and calls it. Unconditional infinite expansion, not a slow document.

It re-arms because biblatex's **apa style asks biber for two sorting schemes**, so the `.bbl` carries two `\datalist` blocks with the same references, and every `\enddatalist` expands to a whole *bare-CS* `\thebibliography … \endthebibliography` (`biblatex_sty.rs::bib_as_thebibliography`). Neither is an environment, so no group, so the first arming was still live when the second `\thebibliography` opened.

The entry's standing suspicion — the `.bbl`'s trailing `\missing{Cowen2021}` — was **wrong**; deleting that line leaves the Fatal untouched.

## The fix — two symmetric halves

- `setup_pseudo_bibitem` captures the originals **once per arming** (`\ifx\bibitem\restoring@bibitem` guard).
- `\endthebibliography` now **disarms**. Upstream has no teardown, relying on `\begin`/`\end{thebibliography}` popping the group the `\let`s were made in — which the bare-CS pair never opens. That leak separately cost a stray empty bibitem outside the biblist (`Error:malformed:ltx:bibitem`) from the blank line after `\printbibliography`. Gated on still being armed, so an `\endthebibliography` reached unarmed cannot copy an undefined `\save@par` onto `\par`.

## Classification: shared with Perl

`\thebibliography \endthebibliography \thebibliography \bibitem{b}` hangs Perl 0.8.8 **>400 s on an 8-line file** that otherwise converts in <2 s. `\ifx\save@bibitem\restoring@bibitem` is true in **both** engines after the second `\thebibliography`.

It stays latent upstream only because Perl's ar5iv biblatex binding never defines `\printbibliography`, so Perl never reads a real `.bbl` this way. Recorded as `KNOWN_PERL_ERRORS.md` **#57**, candidate to upstream.

## Verification

Witness arXiv 2605.17646, same host, ANSI-stripped, Perl run verbose (not `--quiet`):

| | before | after | same-host Perl |
|---|---|---|---|
| outcome | `Fatal:Timeout:TokenLimit` (1e9 tokens), no output | **1 error**, 0.86 s | 59 errors, 33.7 s |
| bibitems / bibliographies / biblists | — | **58 / 2 / 2** | 58 / 2 / 2 ✓ |

The one remaining error is `\missing`, undefined in both engines.

The three shapes the redirection exists to serve all still match Perl element-for-element:

| shape | Rust | Perl |
|---|---|---|
| bibliography with **no** `\bibitem` (blank-line entries) | 3 bibitems / 1 / 1 | 3 / 1 / 1 |
| bare `\thebibliography`, **no** closer | 2 / 1 / 1 | 2 / 1 / 1 |
| `\bibitem` **after** `\end{thebibliography}` | 2 / 2 / 2 | 2 / 2 / 2 |

Gates:
- `cargo test --tests --no-fail-fast`: **1684 passed, 1 failed** — the failure is `latexml_post::graphics::process_coalesces_only_matching_conversion_options`, the known local-only artifact (dev laptop has the full vector toolchain; green in CI), unrelated to bibliographies.
- New guard `cluster_biblatex_two_datalists` verified **red before / green after** (stashed the engine hunk → status_code 3, `Fatal:Timeout:TokenLimit`).
- `cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo doc --workspace` rustdoc-warning-clean.

Blast radius is likely wider than one paper — any apa-style biblatex `.bbl` (two datalists) with a `\par` after `\printbibliography` hits this — but only the single witness was measured, so no fleet number is claimed.

### DIFF
--- a/docs/SYNC_STATUS.md
+++ b/docs/SYNC_STATUS.md
@@ -54,7 +54,7 @@ session of their own. Re-verify a row before planning on it (rule 1).
 | **R1** | Upstream `brucemiller/LaTeXML#2852` — subfile `\documentclass` options | **OPEN upstream**, ours merged as #310 | minutes — chase review, no code | Open items |
 | **R2** | `--preload=<cls>` trips the LaTeX hook stack (`Extra \PopDefaultHookLabel`) | **OPEN**, re-verified 2026-07-25 (1 error with `--preload=article.cls`, 0 without) | small–medium, self-contained | Open items |
 | **R3** | `latexmlmath_oxide` empties a single-structure formula | **OPEN**, re-verified 2026-07-25 (`\frac{1}{2}`→`<mrow/>`; `\frac{a}{b}+c` fine) | small, narrow surface | Open items |
-| **R4** | biblatex `.bbl` `TokenLimit` loop (2605.17646) | real Fatal, pre-existing (bisected — not a PR regression) | medium, needs a dive | Open items |
+| **R4** | biblatex `.bbl` `TokenLimit` loop (2605.17646) | ✅ **FIXED 2026-07-25** — self-referential `\let` on `setupPseudoBibitem` re-arm; shared with Perl | — | Open items |
 | **R5** | Bibliography targets + MakeBibliography re-port | surveyed 2026-07-12; user directive 2026-07-04 | **family** — dedicated session | [`BIBLIOGRAPHY_WORKLIST.md`](parity/BIBLIOGRAPHY_WORKLIST.md) |
 | **R6** | `ltx_env_<name>` env-markup class | user-requested, PLANNED | medium code, **large golden churn** → own branch | Open items |
 | **R7** | Beyond-Perl performance levers BP-1…BP-6 | POST-RELEASE; internal order BP-2 → BP-3 → BP-1 | **family** | [`BEYOND_PERL_LEVERS.md`](performance/BEYOND_PERL_LEVERS.md) |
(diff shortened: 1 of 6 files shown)


dginev:
**Merged `origin/main` (`3b04a8fb79`) in** — main advanced by #377 (R3 `latexmlmath` single-structure formula) and #378 (TL2026 dump-gate docs) while this was open.

**Overlap: one file, `docs/SYNC_STATUS.md`.** No code overlap — #377 touches `bin/latexmlmath_oxide.rs` + a new `005_latexmlmath_single_structure.rs`; this PR touches `latexml_engine/src/latex_constructs.rs` + `06_cluster_regressions.rs`.

Both conflict hunks were the same shape: main deleted the R3 row/section (now fixed) while this branch rewrote the adjacent R4 row/section. Resolution took **both** — main's R3 deletion and this branch's R4 ✅ rewrite.

Audit after resolving:
- `git diff origin/main` over main's five files → **empty** (their side landed byte-identical).
- `git diff` over this branch's files vs the pre-merge commit → **empty** (my side untouched).
- `grep -c R3 docs/SYNC_STATUS.md` → **0**.

**Post-merge full suite: 1686 passed / 92 targets / 0 ignored, 1 failed.** The one red is `latexml_post::graphics::process_coalesces_only_matching_conversion_options` — the known local-only artifact (dev laptop has the full vector toolchain; green in CI here), unrelated to bibliographies. Both sides' guards pass: `single_structure_formula_is_not_emptied ... ok` and `cluster_biblatex_two_datalists ... ok`. Suite-count line in `SYNC_STATUS.md` refreshed to match.

CI green on all four jobs (lint/fmt/clippy/deny, miri, Linux, macOS).

PR #77, opened 2022-03-27
Update to 2022 LaTeXML test suite

dginev:
Copying over the files was the easy part - done. Creaeting a PR here to monitor the diffs as I make headway passing all the tests again.

I will be switching the `.t` files to `.rs` as usual, and marking bits that aren't started as ignored/todo, etc.

Edit: And I was apologizing for a PR with 60 changed files - updating the tests gets us 500+ changed files!

PR #403, opened 2026-07-27
fix(bibliography): a .bib field's specials are data — the two-treatment design

dginev:
Supersedes **#405** (`%`) and **#409** (`&`), which are absorbed here with their
commits, fixtures and guards intact. Rebased onto current `main` (post-#404, #406,
#408).

**Diagnostic.** `% & # _` in a `.bib` field are the characters the author typed,
but they were reaching the TeX tokenizer with live catcodes: `%` (14) commented
out the field's closing brace and took the rest of the entry with it, `&` (4)
was a stray alignment tab and was dropped ("Taylor Francis"), `#` (6) reached
the Stomach as a parameter, `_` (8) was a subscript outside math. Three
separately-reported clusters, one defect.

**Approach — mirror the real dynamic's two treatments.**
`pdflatex → bibtex → pdflatex` treats these characters differently at two
points, and we collapse both into one pass. Restored in order:

1. **Be `bibtex`.** A field's bytes are inert data — bibtex interprets only
   braces and the entry/field delimiters. The read path neutralizes `% & #`
   **without altering the text**: #405's per-Mouth opt-in, generalized from
   percent-only to `Mouth::with_bib_data_literals()`, plus
   `mouth::tokenize_bib_literal` for the handlers that re-read a raw field.
   Per-Mouth and not a State catcode, so a `.sty` raw-loaded from a field
   handler — and the document — keep TeX's meanings.
2. **Be `pdflatex` on the `.bbl` we just synthesized.** Now it must be valid
   TeX, and we know the author meant the literal character, so we escape at that
   boundary: `bibtex.rs::escape_bib_data_specials`, math spans skipped and
   idempotent, at **three** seams — the entry line in `\ProcessBibTeXEntry` plus
   `\bib@@title` and `\bib@@pages`, which re-read the RAW field rather than the
   value they were handed, so escaping only the entry line missed every title.

**`_` is in treatment 2 only.** A catcode is fixed at tokenization and cannot
tell whether it is inside `$…$`, and a subscript in a title's math is legitimate
TeX. Measured: putting `_` in the Mouth set silently flattened every subscript
in a bibliography title. Only the escaper can be math-aware.

**The exclusion list is the same insight, not an ad-hoc list.** A handler that
consumes the field's characters *itself*, under its own catcode regime — `url`'s
Verbatim href, `doi`'s id — is still operating in treatment 1, on data, so it
receives the value **unescaped**. `bib_field_source` reads that declaration off
the `Definition`. Consolidated as `OXIDIZED_DESIGN #74`, superseding #405's
percent-only and #409's ampersand-only entries; #409's
`undouble_escaped_ampersand` is kept as the separate input corruption it is.

**Validation.** Suite **1721 passed / 1 failed** (101 targets after #404) — the pre-existing
(issue 401). `clippy --workspace --all-targets -D warnings` and `fmt --check`
clean. All inherited canaries pass, including
`bib_name_space_form_accent_survives_reversion` (PR #399),
`bib_field_percent_is_an_ordinary_character`,
`bib_bare_ampersand_is_literal_data`,
`bib_bare_ampersand_leaves_live_markup_alone`,
`bib_escaped_amp_entity_decodes_to_one_ampersand` and
`55_bibtex::runaway_field_costs_only_its_own_entry`.

New: `bib_field_specials_are_data_not_tex` over a ten-entry fixture proving
together — the four specials bare render literally; the already-escaped entry
renders an **identical** title (no `\\&`); `$x_1+x_2$` keeps its `SUBSCRIPTOP`;
`\emph` still produces markup; `{\v S}pakov`/`Gon{\c c}alves` still revert;
`%20` inside `\url{…}` survives; `url`/`doi` pick up no backslash. Plus six
`escape_bib_data_specials` unit tests isolating the `\\&` hazard.

**All sixteen witnesses, all three clusters**, `--release` before/after on the
same host, TOTAL document errors:

| cluster | witness | before | after |
|---|---|---|---|
| `_` | 2605.06926 | 8 | **0** |
| `_` | 2605.01936 | 13 | **0** |
| `_` | 2605.04604 | 2 | **0** |
| `_` | 2605.08986 | 2 | **0** |
| `_` | 2605.05898 | 1 | **0** |
| `_` | 2605.11300 | 1 | **0** |
| `_` | 2605.11579 | 5 | 5 † |
| `%` | 2605.01196 | 28 | **0** |
| `%` | 2605.02131 | 28 | **0** |
| `%` | 2605.00879 | 103 | **1** ‡ |
| `&` | 2605.06249 | 3 | **0** |
| `&` | 2605.03054 | 1 | **0** |
| `&` | 2605.00462 | 1 | **0** |
| `&` | 2605.08753 | 1 | **0** |
| `&` | 2605.10409 | 1 | **0** |
| `&` | 2605.00833 | 0 | **0** § |

**193 → 0.** No number is worse than the separate PRs achieved, and two are
better: **2605.01936 13 → 0** where #409 alone reached 6 (its `_` half is now
fixed too), and **2605.06624 4 → 0** where #409 measured 3.

† 2605.11579 has **zero** `unexpected:_` before AND after — it never belonged to
the `_` cluster. Its apparent three were an artifact of measuring in the
DEGRADED no-dump mode a fresh worktree starts in.
‡ the residual is `undefined:\mathsemicolon`, unrelated and unchanged.
§ a RENDERING witness, not an error-count one: its `\&amp;` printed "&amp;" and
now prints "&" (`bib_escaped_amp_entity_decodes_to_one_ampersand`).

## Decisions & open questions

**1. What I dropped from the three branches.** From my own earlier work: the
`eprint`/`preprint`/`archive` **`Semiverbatim` change is gone**, and its
divergence entry with it — treatments 1+2 handle those fields, so it no longer
earned its place, and the bindings are back to Perl's exact shape. Its fixture
entries survive inside `bib_field_specials.bib`, and its guard was folded into
the consolidated one. From #405/#409: nothing dropped; `with_percent_as_other` /
`with_align_as_other` were merged into one `with_bib_data_literals`, and their
two divergence entries into #74.

**2. Doing BOTH treatments is belt-and-braces, deliberately.** For a value that
goes through the entry line they overlap — the escaper makes it valid TeX and
the Mouth would also have neutralized it. I kept both because they cover
different populations: the Mouth catches what the escaper deliberately leaves
alone (a `url` href, a `doi` id, the inside of `\url{…}`), and the escaper
covers `_`, which the Mouth cannot. **Alternative:** treatment 2 only, with a
math-aware pre-pass everywhere. **What would settle it:** whether any value
reaches a tokenizer by a fourth route neither covers.

**3. `^` is deliberately outside the set.** So `x_1` in a note now renders
literally while `x^1` still raises `unexpected:^`. That asymmetry is real. It is
not in the authorized four, outside math it is essentially always a typo, and it
is the live probe in `105_bib_field_digest_once` (whose `_` probe this work
retired — that test now asserts `_` contributes zero and keeps `^` as the
once-only probe). **What would settle it:** a corpus count of `^` outside math
in `.bib` fields.

**4. An `&` inside a field's `$\begin{array}…$` would lose its alignment.**
Treatment 1 is catcode-level and cannot be math-aware, and `&` is in treatment 1
because #409 measured and shipped it there. No witness exhibits one. **What
would settle it:** moving `&` to treatment 2 as well, which the escaper's
math-span skip would handle — deferred rather than done, because it would
re-open a measured, guarded decision on no evidence.

**5. Escaping runs BEFORE `recase_title`.** That pass splits on words and treats
a `\…` escape as part of its word, so raw `AT&T` would recase to `AT&t` while
`AT\&T` recases to `AT&T`; escaping last made the two spellings render
*differently*, which is the invariant the guard exists to protect. Visible
casing change for bare-`&` titles versus pre-fix (which was an error, not a
rendering). **What would settle it:** what `bibtex(1)`'s `change.case$` does to
`AT&T` — reasoned from `plain.bst`, not executed.

**6. Two previously-green guards changed behaviour, deliberately.**
`55_bibtex::runaway_field_costs_only_its_own_entry` (inherited from #405) now
expects the `%`-bearing title WHOLE where it expected Perl's truncation;
`105_bib_field_digest_once` swapped its probe as in (3). Flagging both rather
than letting them read as incidental edits.

**6b. One correction to the landed #408 doc, and the sites it named.** #408's
table puts `_` in **treatment 1**. Measured, that is wrong: a catcode is decided
at tokenization, before anything knows whether the character is inside `$…$`,
and putting `_` in `Mouth::with_bib_data_literals` silently flattened every
subscript in a bibliography title — caught by #409's own
`bib_bare_ampersand_leaves_live_markup_alone`, which asserts `role="SUBSCRIPTOP"`.
So treatment 1 covers `% & #` and treatment 2 covers all four. The table row is
corrected in place in `BIBLIOGRAPHY_WORKLIST.md` with the measurement and the
guard name. Separately, #408 says any change "must cover all three [seams], plus
the name-, date- and MR/Zbl-assembly sites that share that path" — those sites
had treatment 1 only, so treatment 2 now reaches them too, through one
`bib_tex_tokens` helper (escape, then tokenize-as-data). Escaping is idempotent,
so the title and pages sites that already escaped upstream are unaffected.

**7. Divergence numbering.** One consolidated `#74`; `#75` stays #406's;
`#76` retired with the `Semiverbatim` change. Please re-check at merge — the
numbers moved three times while this was in flight.

**8. Process note.** `git stash` is shared across worktrees and cost two
cross-agent collisions during this work (a sibling's stashed `bibtex.rs` landed
in my tree and mine in theirs; both recovered). Use `git diff > patch` +
`git checkout --` instead. Separately, a fresh worktree has no
`resources/dumps/`, so its first suite run reports ~6 unrelated failures from
DEGRADED raw-load mode and its first error counts are inflated — run
`tools/make_formats.sh` before believing any measurement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,12 @@
     a percent that stays literal (BibTeX has no comments, and `%` is routine in
     an encoded URL). Measured over a 30,079-document arXiv sample: 62 more
     documents convert cleanly, 44 fewer carry errors.
+  - **An ampersand in a reference is an ampersand** — "Taylor & Francis" in a
+    `.bib` publisher, journal or booktitle used to report an error and print
+    "Taylor Francis". A `.bib` field's content is data, so the character is kept.
+  - **A reference exported through HTML no longer shows `&amp;` for `&`** — a
+    doubly escaped ampersand (`\&amp;`) in a `.bib` title, journal or booktitle
+    is decoded back to the one character the author wrote.
   - **amsrefs `\bib` values digest as live TeX** — `\MR{…}` came out as literal
     characters and `pages` rendered empty.
   - **MathReview / ZentralBlatt links are synthesized** — `mrnumber`/`zblno`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,9 +74,12 @@
     a percent that stays literal (BibTeX has no comments, and `%` is routine in
     an encoded URL). Measured over a 30,079-document arXiv sample: 62 more
     documents convert cleanly, 44 fewer carry errors.
-  - **An ampersand in a reference is an ampersand** — "Taylor & Francis" in a
-    `.bib` publisher, journal or booktitle used to report an error and print
-    "Taylor Francis". A `.bib` field's content is data, so the character is kept.
+  - **A reference's `& % # _` are the characters the author typed** — "Taylor &
+    Francis" in a publisher used to report an error and print "Taylor Francis",
+    a gene id `AT1G01010_v2` came out as a subscript error, and a percent in an
+    encoded URL commented out the rest of the entry. A `.bib` field's content is
+    data, so all four are kept, while `\emph{…}`, `$x_1+x_2$` and accents in the
+    same field keep working.
   - **A reference exported through HTML no longer shows `&amp;` for `&`** — a
     doubly escaped ampersand (`\&amp;`) in a `.bib` title, journal or booktitle
     is decoded back to the one character the author wrote.
--- a/docs/SYNC_STATUS.md
+++ b/docs/SYNC_STATUS.md
@@ -64,6 +64,7 @@ session of their own. Re-verify a row before planning on it (rule 1).
 
 ## Current status
 
+- **2026-07-26 (later still) — a bare `&` in a `.bib` field is data (OXIDIZED_DESIGN #74).**
 - **2026-07-26 — undefined CSes from packages with no binding: `silence.sty`,
   bundled `arxiv.sty`/`PRIMEarxiv.sty`.** Long-standing gaps, **not** a
   regression: Perl 0.8.8 has no binding for either and reproduces the identical
@@ -89,15 +90,17 @@ session of their own. Re-verify a row before planning on it (rule 1).
   `.bbl` under `plain` and `abbrvnat`; pdflatex stops with "Misplaced alignment
   tab character &" and prints "Taylor Francis"). **A `.bib` field's content is
   DATA** — authorized surpass-Perl and surpass-pdflatex, since LaTeXML reads
-  `.bib` directly and decides what reaches the tokenizer. Neutralized at the
-  same two seams #74 uses for `%`, and BOTH are required — the per-entry Mouth
-  (`Mouth::with_align_as_other`) and `mouth::tokenize_bib_literal` for handlers
-  that re-tokenize a stored raw field (`\bib@@title` recasing, name/date/pages
-  assembly, MR/Zbl). Measured: **2605.06249 3→0, 2605.03054 1→0, 2605.00462 1→0,
-  2605.08753 1→0, 2605.10409 1→0, 2605.01936 13→6, 2605.06624 4→3** (the
-  residuals are unrelated `undefined:` errors). `#` was checked and NOT included
-  — one bare `#` across all seven, in a JabRef `file` path already covered as an
-  unknown field. **Stacked on PR #405**: merge after it.
+  `.bib` directly and decides what reaches the tokenizer.
+  Landed inside the consolidated **OXIDIZED_DESIGN #74**, which covers `%`, `&`,
+  `#` and `_` under one two-treatment design — be `bibtex` (the per-entry Mouth
+  and `mouth::tokenize_bib_literal`, via `Mouth::with_bib_data_literals`), then
+  be `pdflatex` on the `.bbl` you just synthesized
+  (`bibtex.rs::escape_bib_data_specials`, at three seams: the entry line,
+  `\bib@@title` and `\bib@@pages`). `_` is in the escaper ONLY: a catcode is
+  fixed at tokenization and cannot tell whether it is inside `$…$`, and a
+  subscript in a title's math is legitimate TeX — putting `_` in the Mouth set
+  flattened every one of them. Measured across all sixteen witnesses of the
+  three clusters (`_`, `%`, `&`): **193 -> 0**.
   Also fixed, a different bug the neutralization does *not* reach: the doubly
   escaped `\&amp;` / `{\&}amp;` / `&amp;`, an HTML entity that survived into
   the `.bib` and printed as "&amp;" in Perl and pdflatex alike
--- a/docs/SYNC_STATUS.md
+++ b/docs/SYNC_STATUS.md
@@ -81,6 +81,31 @@ session of their own. Re-verify a row before planning on it (rule 1).
   1→0, 4→1, 1→0, 1→0. Divergence #77. Guards `00_contrib::{silence_filters,
   arxiv_keywords, primearxiv_keywords}_test`, `106_arxiv_sty_defers_to_bundled`,
   `107_silence_keeps_diagnostics`.
+- **2026-07-26 (later still) — a bare `&` in a `.bib` field is data (OXIDIZED_DESIGN #75).**
+  Seven 2605 witnesses carried `Error:unexpected:&` from `publisher` / `journal`
+  / `booktitle` / `author` / `copyright` ("Taylor & Francis"). Not a Rust-only
+  defect: same-host `latexmlc` raised the identical per-`&` count on all six
+  re-measured witnesses, and bibtex 0.99d + pdflatex agree (the `&` reaches the
+  `.bbl` under `plain` and `abbrvnat`; pdflatex stops with "Misplaced alignment
+  tab character &" and prints "Taylor Francis"). **A `.bib` field's content is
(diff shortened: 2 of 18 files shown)


dginev:
**Pre-merge verification on the rebase onto current `main`** (which now also carries #404, #406, #408). Checked both halves — error count *and* whether the bibliography is actually populated, since an error-free run with a dropped bibliography would look like success.

**14 of 16 witnesses: 0 errors, populated bibliography, zero `ltx_ERROR` spans.**

| witness | err | bibitems | witness | err | bibitems |
|---|---|---|---|---|---|
| 2605.00462 | 0 | 19 | 2605.04604 | 0 | 98 |
| 2605.00833 | 0 | 29 | 2605.05898 | 0 | 17 |
| 2605.01196 | 0 | 83 | 2605.06249 | 0 | 45 |
| 2605.01936 | 0 | 39 | 2605.08753 | 0 | 62 |
| 2605.02131 | 0 | 20 | 2605.08986 | 0 | 91 |
| 2605.03054 | 0 | 11 | 2605.10409 | 0 | 42 |
| 2605.00879 | 1 | 150 | 2605.11300 | 0 | 78 |

**One number in the PR body needs softening.** `2605.06926` is reported as `8 → 0`; measured here it is **4 errors and 0 bibitems**. The `_` cluster *is* fully fixed — all eight `unexpected:_` are gone — but the paper is `\usepackage[style=authoryear]{biblatex}` + `\printbibliography`, and the residue is `undefined:\xpatchbibdriver` / `\usebibmacro` / `\newunit` with an empty bibliography. That is the standing biblatex limitation, untouched by this PR. The cluster claim holds; the whole-document claim does not.

(My first pass also showed `Fatal:Timeout:Convert` on this witness — that was my own artifact, a debug build under a 400s cap. Re-run at 1800s: no timeout, same 4 errors.)

**`2605.11579` is correctly disclaimed.** Its three `unexpected:_` are at `main; line 31` — the document body, not `Anonymous String`. Confirmed not a `.bib` defect. Worth flagging separately though: it has 0 bibitems and five undefined CSes (`\io`, `\ic`, `\Sec`, `\cprimefand`, `\ccois`), and it *was* `no_problem` in cortex run 269 — so something else regressed it. Not this PR's scope; filing follow-up.

Numbering after rebase is consistent: **#74** consolidated, **#75** = #406, **#77** = #404, **#76** now a deliberate gap since the `eprint` `Semiverbatim` entry was dropped as redundant. All `.rs`/`.md` references resolve.

dginev:
**Correction to my previous comment, and a stricter re-verification.**

Re-ran all 16 against the **emitted `.latexml.log`** (`--log=`) rather than captured stderr — that log is ANSI-free by construction and is the canonical signal. Both signals agree, and one of my numbers was wrong.

| witness | errors *in the emitted log* | bibitems | | witness | errors | bibitems |
|---|---|---|---|---|---|---|
| 2605.00462 | 0 | 19 | | 2605.06249 | 0 | 45 |
| 2605.00833 | 0 | 29 | | 2605.08753 | 0 | 62 |
| 2605.01196 | 0 | 83 | | 2605.08986 | 0 | 91 |
| 2605.01936 | 0 | 39 | | 2605.10409 | 0 | 42 |
| 2605.02131 | 0 | 20 | | 2605.11300 | 0 | 78 |
| 2605.03054 | 0 | 11 | | 2605.05898 | 0 | 17 |
| 2605.04604 | 0 | 98 | | | | |

**13 witnesses: zero `Error:`/`Fatal:` lines in the emitted log, populated bibliography.** (Warnings range 0–12 and are not in scope.)

**`2605.06926` — I was wrong.** I reported "4 errors and 0 bibitems". Measured cleanly, on an idle box, with the emitted log: **3 errors and 89 bibitems**. The bibliography renders fine; the residue is only `undefined:\xpatchbibdriver` / `\usebibmacro` / `\newunit`, the standing biblatex gap. My earlier figure came from a run launched while 16 parallel conversions were still finishing — contention truncated the output, and the "0 bibitems" and the extra error were both artifacts of that, as was the `Fatal:Timeout` I saw on the pass before. Three measurements of the same witness, three different numbers, all from my harness rather than the code. The PR body's `8 → 0` is still worth softening to name the biblatex residue, but the witness is in far better shape than I said.

**`2605.00879`** — 1 error (`undefined:\mathsemicolon`, unrelated), 150 bibitems. As the PR states.

**`2605.11579`** — 9 errors, 0 bibitems against 889 `@` entries. Confirmed **not** this cluster: its `unexpected:_` are at `main; line 31`, the document body. Filed separately as issue 410.

Net: **15 of 16 witnesses have a populated bibliography**, and 13 are completely clean in the emitted log.

PR #405, opened 2026-07-27
fix(bibliography): a percent in a .bib field is data, not a comment

dginev:
**Diagnostic.** Both witnesses are one root cause: a `%` inside a `.bib` field
value. `Pre::BibTeX` is a faithful BibTeX parser — `%` is significant only in
the junk BETWEEN entries (`skipJunk`), so the stored value keeps it — but
`\bibentry@create` re-injects that value as TeX source, one
`\csname bib@field@…\endcsname{<value>}` per line, where catcode 14 comments
out the rest of the line **including the field's own closing brace**. The entry
group never closes, the element it opened is left dangling, and every later
entry nests inside it. That is the `malformed:ltx:bibentry` half; the
`expected:}` and `unexpected:\end{bibtex@bibliography}` are the same event seen
at the gullet. The "Currently in …" breadcrumb names the dangling element
differently per witness — `<ltx:bibentry>` for `2605.01196`
(`doi={%doi:10.1017/jfm.2016.420}`, an author's note-to-self in an *uncited*
entry) and `<ltx:bib-title>` for `2605.02131` (a percent-encoded URL in a
`title`'s `\href`) — because the two fields reach the tokenizer by different
routes. One-step bisect confirmed it: neutralize that single character in the
`.bib` and both go 28 → 0.

**Approach.** Read the text BibTeX lexed with `%` as an ordinary character, at
both seams:

* `Mouth::with_percent_as_other()` on the per-entry mouth. A **per-Mouth**
  property, not a `\catcode` assignment — a State catcode would be inherited by
  a `.sty` raw-load triggered from inside a field handler, where `%` must stay a
  comment. Opt-in, default off; every existing mouth is untouched.
* `mouth::tokenize_percent_literal` behind `bibtex.rs::tokenize_bib_field`, for
  the handlers that re-tokenize a stored **raw** field and never touch that
  mouth: `\bib@@title` recasing, name splitting, date/pages assembly, MR/Zbl.

Ground truth is `bibtex` 0.99d + pdflatex, run on the fixture: bibtex **drops**
`doi` (not in `plain.bst`'s `ENTRY` list, so pdflatex never sees `%doi:`) and
**keeps** the `%` verbatim in a `title`, where pdflatex compiles it **cleanly**
because real hyperref's `\href` sets `` \catcode`\%=12 `` before reading its
URL. Perl LaTeXML breaks on both — so this is surpass-Perl on a shared bug,
recorded as `OXIDIZED_DESIGN #74` (sibling of #73, same "opened and never
closed" shape).

**Validation.** Guard
`06_cluster_bibliography::bib_field_percent_is_an_ordinary_character`
(`convert_and_post_clean` — the bibliography is built in POST, and a
text-presence assertion still finds text inside an unclosed element). RED
verified by reverting the two source files: 7 errors, the exact cluster
signature including both breadcrumb shapes. The fixture carries one entry per
seam, a containment canary, and an entry holding `\emph`, inline math and a
space-form accent **alongside** the `%`, so the neutralization cannot quietly
flatten a field.

| witness | before | after | same-host Perl `latexmlc` |
|---|---|---|---|
| `2605.01196` | 28 | **0** | 29 |
| `2605.02131` | 28 | **0** | 31 |
| `2605.00879` (found by the sweep below) | 103 | **5** | — |

`2605.01196` output is otherwise byte-identical (83 bibitems before and after —
the entire cost was diagnostics for a reference nobody cites). `2605.02131`
keeps its 20 bibitems and now renders the two `\href` titles, URLs intact.
`2605.00879` is `note = {https://doi.org/10.1145%2F3292500.3330925}`; its
residual 5 belong to other clusters (`unexpected:_` ×3, `\mathsemicolon`, and
one `ltx:bibitem` in `<ltx:bib-title>`) — i.e. the two sibling agents' clusters,
cleanly separated from this one.

**Corpus reach.** Over the first 1200 papers of sandbox-arxiv-2605, 178 ship a
`.bib` with a `%` somewhere in a field value; 66 have an *unescaped* `%` in a
field other than the three #73 already reads `Verbatim`. Of the 15 whose field
is one that is neither `Semiverbatim` nor `Verbatim` (`doi`, `note`, `journal`,
`howpublished`, `adsurl`), **two were broken and are now clean** (the two rows
above) and the other 13 were already fine — `url` (#72) and unknown fields
(`\bib@field@default@default Verbatim Verbatim`) had their catcodes protected by
their parameter type. So this closes exactly the population no parameter type
was covering: the *rendered* fields.

Suite `1705 passed / 1 failed`, the one failure being the pre-existing
(issue 401, red on clean `main` on this host). `clippy -D warnings` and
`fmt --all` clean.

## Decisions & open questions

1. **This overlaps the shared `.bib` specials mechanism — probably drop mine.**
   The maintainer note says a sibling agent owns neutralizing `_ & # %` in the
   `.bib` reading path, and that a `%`-eats-the-closing-brace cause belongs to
   that mechanism. My cluster turned out to be **entirely** that cause: there is
   no separate structural half here. The `malformed:ltx:bibentry` errors are
   strictly downstream — killing the one `%` takes them to zero with no residue,
   so there is no "element opened on empty content" defect of the #73 kind to
   fix independently. I had already implemented and validated this before the
   note arrived; I am shipping it rather than stalling, but **if the sibling's
   mechanism lands first, close this in favour of it** and keep only the fixture
   + guard, which should pass unchanged under either implementation.
   *Handover intel the sibling will need:* neutralizing at the reading path
   **alone is not sufficient**. `\bib@@title` bypasses the entry mouth
   completely — it re-reads the RAW stored field to re-case it and tokenizes the
   result itself under the standard cattable. Fixing only the mouth took witness
   `2605.02131` from 28 to **37** errors (an `Extra \endcsname` surfaced once the
   value stopped being truncated). Any shared mechanism must cover
   `bibtex.rs::current_entry_field` / `\bib@@title` / the name-, date-, pages-
   and MR/Zbl-assembly call sites, not just the `.bib` read.
   *Evidence that would settle it:* whether the sibling's neutralization is
   applied to the stored value or to the tokenizer — if the former, both seams
   are covered and mine is redundant.

2. **Touching shared machinery: `latexml_core::mouth`.** `get_next_char` is on
   the hot path for every mouth in the workspace. The conservative shape was
   chosen: one `bool` field defaulting to `false`, one builder, and the catcode
   lookup factored into a `catcode_of` helper that is a pure passthrough unless
   the flag is set — so behaviour is bit-identical for every existing caller,
   and only comment-ness is removed (a `%` given some other catcode keeps it;
   `\%` is unaffected, since a control symbol is formed from the following
   character whatever its catcode). *Alternative considered and rejected:*
   injecting `` \catcode`\%=12 `` into the generated entry TeX — simpler and
   needs no core change, but it is a State assignment, so a raw `.sty` loaded
   from inside a field handler would inherit it and lose its own comments.
   *Evidence that would settle it:* if the sibling's mechanism works purely on
   the stored string, the core change can be dropped entirely.

3. **Did NOT read `doi`/`title` verbatim the way #73 read
   `abstract`/`keywords`/`contents`.** #73 could afford `Verbatim` because
   nothing renders `ltx:bib-extract`; `doi` and `title` are rendered, so
   verbatim would destroy their markup — and the boundary requires `\emph`,
   math and accents to keep working. Fixing the *reading* of `%` instead
   preserves everything. This is why the fixture asserts markup/math/accent
   survival in a percent-bearing field.

4. **Realigned a pre-existing guard whose premise I invalidated.**
   `55_bibtex::runaway_field_costs_only_its_own_entry` explicitly asserted the
   `%` truncation as parity ("deliberately NOT corrected"). It now asserts the
   values survive whole. Its containment half needed a live runaway, so the
   fixture gains an entry with an unescaped `$5` in a `note` (a common `.bib`
   malformation) which still digests to EOF. Measured and stated honestly in the
   test doc: with `%` retired as a trigger, **no `.bib` input reachable through
   `Pre::BibTeX` produces a token-level unbalanced read any more** (the parser
   balances braces character-wise before we see the value), so the entry mouth's
   `BalancedBoundary::Opaque` is now an invariant the fixture states rather than
   one it can provoke — I verified this by flipping it to `Transparent` and
   getting identical output. Flagging it rather than deleting the boundary.

5. **No CHANGELOG entry.** The existing 0.7.5 bibliography bullet already says
   "a percent that stays literal (BibTeX has no comments, and `%` is routine in
   an encoded URL)" from the #72 work, and the sibling's shared-mechanism PR is
   the natural owner of the user-facing line for `_ & # %` as a family. Skipped
   to avoid a duplicate bullet and a near-certain conflict. Say the word and I
   will add one.

6. **Scope held to the cluster.** Left unfixed: `\href` is *undefined* in the
   recursive bib session when the article loads neither `hyperref` nor a package
   providing it, so a `.bib` `title` using `\href` raises
   `Error:undefined:\href` (with the `%` fixed, that is now the only remaining
   error on such a document). That is divergence #72's territory — #72 loads
   `url.sty` when `\url` is undefined and could plausibly do the same for
   `\href` — and neither witness needs it (both have `\href` defined). Not
   touched.

### DIFF
--- a/docs/parity/OXIDIZED_DESIGN_DIVERGENCES.md
+++ b/docs/parity/OXIDIZED_DESIGN_DIVERGENCES.md
@@ -2275,6 +2275,76 @@ this fixture, 203 on the witness) passed every bibliography guard silently.
 (percent in `abstract`, specials in `keywords`, and a third entry as the
 containment canary).
 
+### 74. A `%` inside a `.bib` field value is data, not a comment
+
+**Perl behaviour.** `Pre::BibTeX` is a real BibTeX-format parser: field values
+come out of `parseString`/`parseBalancedBraces` as raw strings, and `%` is
+significant **only** in the junk between entries (`skipJunk`, `BibTeX.pm` L335-343
+— its own comment reads "Although % officially starts comments, apparently BibTeX
+accepts anything until @"). Inside an entry there is no comment syntax, matching
+`bibtex.web`. `BibTeX.pool.ltxml` L134-166 (`\bibentry@create`) then re-serializes
+those values into TeX source — one `\csname bib@field@<t>@<f>\endcsname{<value>}`
+per line — and hands the join to `Mouth->new($tex)`. There `%` is catcode 14
+again, so it comments out the rest of its line, **the field's own closing brace
+included**. The entry's group never closes, `<ltx:bibentry>` is left open, and
+every following entry nests inside it: `<ltx:bibentry> isn't allowed in
+<ltx:bibentry>`, once per remaining entry. `\bib@@title` (L293-333) loses the
+value a second way — it re-reads the RAW field to re-case it and `Tokenize`s the
+result itself, and `Tokenize` uses the standard cattable, where `%` is again a
+comment; the truncated title leaves a `\href` mid-argument and the runaway eats
+the following `\csname`s (`Extra \endcsname`).
+
+**Rust behaviour.** The text BibTeX lexed is read with `%` as an ordinary
+character. Two seams, because the value reaches the tokenizer two ways:
+`Mouth::with_percent_as_other()` on the per-entry mouth
+(`\ProcessBibTeXEntry`, `bibtex.rs`), and `mouth::tokenize_percent_literal`
+behind `bibtex.rs::tokenize_bib_field` for the handlers that re-tokenize a
+stored raw field (title recasing, name splitting, date/pages assembly, MR/Zbl).
+It is a per-**Mouth** property rather than a `\catcode` assignment on purpose: a
+State catcode would be inherited by a `.sty` raw-load triggered from inside a
+field handler, where `%` must stay a comment. Only comment-ness is removed — a
+`%` given some other catcode keeps it, and `\%` is unaffected (a control symbol
+is formed from the following character whatever its catcode).
+
+**Why — measured against `bibtex` 0.99d + pdflatex, not against Perl.** The
+fixture's three entries were run through the real toolchain (TL2025,
+`plain.bst`, hyperref loaded), and it handles both fields cleanly:
+
+* `doi` — **`bibtex` drops it.** `plain.bst` does not declare `doi` in its
+  `ENTRY` list, so the generated `d.bbl` carries author/title/journal/year and
+  no trace of `%doi:`. pdflatex never sees the character. (Nor would it in the
+  witness: 2605.01196's entry sits in that file's own "UNUSED REFERENCES" block,
+  which `bibtex` never processes at all.)
+* `title` — **`bibtex` keeps the `%` verbatim** (its lexer has no comment rule
+  inside `{}`), writing `\href{https://…/20240723%20IPWG%20…DRAFT.pdf}{A linked
+  title…}` into the `.bbl` with the closing brace on that same line. pdflatex
+  **compiles it, rc=0**, and the PDF renders the linked title: real hyperref's
+  `\href` sets `` \catcode`\%=12 `` before reading its URL. So the ground-truth
+  engine reads a percent-encoded URL with `%` as an ordinary character — the
+  exact rule adopted here.
+
+LaTeXML's `HyperVerbatim` does the same catcode trick as hyperref, but far too
+late: the `%` is eaten during the *field* read, before `\href` is ever reached.
+Reading the `.bib` **directly** — which is what both LaTeXML engines do, and
+what no other consumer does — is what puts the value in front of a TeX tokenizer
+in the first place, so the reader has to use BibTeX's lexing rules for it. The
+damage under the old reading is also wildly disproportionate to the input: one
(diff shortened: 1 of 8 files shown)


dginev:
Superseded by #403, which absorbed this branch's commits, fixtures and guards **verbatim** and landed as `966b3b2de6`.

Nothing here is discarded. The joint two-treatment design keeps this PR's mechanism in its correct place — treatment 1 (reading `.bib`, be bibtex) neutralizes `% & #` at the mouth without altering the text; treatment 2 (synthesizing the `.bbl`, be pdflatex pass 2) escapes at the data→TeX boundary and covers `_` as well, since only the escaper can be math-aware.

The consolidated result is better than this PR reached alone: **193 → 0** across all 16 witnesses of the `_`/`%`/`&` clusters, with 2605.01936 going 13→0 (this branch reached 6) and 2605.06624 4→0 (measured 3 here). Recorded as OXIDIZED_DESIGN #74.

PR #409, opened 2026-07-27
fix(bibliography): an ampersand in a .bib field is an ampersand

dginev:
> **Stacked on #405** — the first commit here is that PR's. Merge #405 first;
> this rebases onto it and its diff then reduces to the `&` half.

**Diagnostic.** Seven arXiv/2605 papers report `Error:unexpected:&` at
`Anonymous String` — the recursive BibTeX session. The `&` is real `.bib`
content: `publisher = {Taylor & Francis}`, `journal = {Infrared Physics &
Technology}`, `booktitle = {… Knowledge Discovery & Data Mining}`, `author = {…
Daishi Kuroiwa & Matteo Rocca}`, `copyright = {… Lippincott Williams &
Wilkins}`. BibTeX's lexer has no alignment, so that `&` is a character in a
name; TeX reads catcode 4, errors, and **drops** it — the entry printed "Taylor
Francis". This is *not* a Rust-only defect: same-host `latexmlc` raised the
identical per-`&` count on all six re-measured witnesses (1/1, 3/3, 1/1, 1/1,
1/1, 1/1), and `bibtex` 0.99d + pdflatex agree — under `plain` and under
natbib's `abbrvnat` the bare `&` is copied into the `.bbl` and pdflatex stops
with `! Misplaced alignment tab character &`, printing "Taylor Francis". Rust
already beat Perl on these overall (2605.01936: 13 errors and a complete
bibliography vs Perl's 101 + a Fatal + none at all).

A second, unrelated ampersand bug turned up in the same sweep: `\&amp;` — an
HTML entity that survived into the `.bib` and was TeX-escaped a second time.
That one raises no error anywhere (Perl and pdflatex both print "&amp;"), so the
catcode work does not reach it.

**Approach.** A `.bib` field's content is **data, not TeX** (maintainer policy)
— authorized surpass-Perl and surpass-pdflatex, because LaTeXML reads `.bib`
directly, with no `.bst` and no `bibtex(1)`, and so decides what reaches the
tokenizer. `&` is regime A, beside #405's `%`: it derails alignment in *any*
field, and these fields have no Perl `Semiverbatim` precedent to follow (unlike
`doi`/`isbn`/… which `eprint` joins in #403). Both of #405's seams are extended,
and both are needed — the per-entry Mouth (`Mouth::with_align_as_other`) and
`mouth::tokenize_bib_literal` for the handlers that re-tokenize a stored raw
field and never pass through it (`\bib@@title` recasing, name/date/pages
assembly, MR/Zbl). Per-Mouth, not a State catcode, so a `.sty` raw-loaded from
inside a field handler — and the document — keep TeX's alignment tab.
`undouble_escaped_ampersand` handles the `\&amp;` family separately, at the
entry-assembly seam and again in `\bib@@title`. `OXIDIZED_DESIGN #75`.

**Validation.** Per-witness `Error:` totals, debug profile with dumps, baseline
= #405 tip:

| witness | before | after | `unexpected:&` |
|---|---|---|---|
| 2605.06249 | 3 | **0** | 3 → 0 |
| 2605.03054 | 1 | **0** | 1 → 0 |
| 2605.00462 | 1 | **0** | 1 → 0 |
| 2605.08753 | 1 | **0** | 1 → 0 |
| 2605.10409 | 1 | **0** | 1 → 0 |
| 2605.01936 | 13 | **6** | 7 → 0 |
| 2605.06624 | 4 | **3** | 1 → 0 |

No `unexpected:&` remains on any of the seven; the two residuals are unrelated
`undefined:` errors. #405's own canaries are unaffected (2605.02131 and
2605.01196: 0 errors both before and after this change). Entity witness
2605.00833 renders "Computer Engineering, & Applied Computing" (was `&amp;`),
0 errors.

Guards, each RED-verified by reverting the relevant hunk (`git diff` +
`git checkout`, never `git stash` — that stack is shared across worktrees on
this box and it swapped two agents' trees today):
`06_cluster_bibliography::bib_bare_ampersand_is_literal_data` (5 post-stage
errors → 0, and "Taylor & Francis" instead of "Taylor Francis"),
`::bib_bare_ampersand_leaves_live_markup_alone`, and
`::bib_escaped_amp_entity_decodes_to_one_ampersand` (`&amp;amp;` → `&amp;`).
All three gate the POST stage with `convert_and_post_clean`. Clippy and fmt
clean; full suite green apart from the known
(issue 401, fails on clean main on this host).

## Decisions & open questions

1. **Where to neutralize `&` — per-Mouth (regime A) vs per-field `Semiverbatim`
   (regime B).** Chose regime A, per the blast-radius rule: `&` is destructive
   in any field, and `publisher`/`journal`/`booktitle` have no Perl per-field
   idiom to mirror. Regime B would have meant inventing `Semiverbatim` for five
   ordinary rendered fields, which also suppresses `~ $ ^ _ '` in them — far
   more than the evidence supports. *What would settle it:* a witness needing
   `&` literal but something else in the SPECIALS set live.
2. **Both seams, or just the mouth.** Both. The `%` agent measured 2605.02131
   getting *worse* (28 → 37) with the mouth alone, because `\bib@@title` re-reads
   the raw field. Confirmed directly here: with only the mouth covered, `\&amp;`
   in a `title` stayed broken while the same shape in a `booktitle` was fixed.
3. **`#` deliberately not included.** Checked rather than assumed: exactly one
   bare `#` across all seven witnesses, in 2605.00462's JabRef `file` path, and
   that field is already covered (unknown field → `Verbatim Verbatim`).
   *Settles it:* a witness with a bare `#` in a rendered field.
4. **`\&amp;` — repair, or render faithfully?** Chose repair. pdflatex is the
   named tiebreaker and pdflatex prints `&amp;`, but it has no way to know
   better; reading the `.bib` we can see one file mixing two escaping
   conventions, and the entity is unambiguous — no author writes `\&amp;`
   meaning six characters. The alternative ("garbage in, garbage out") is
   defensible; it just means every reader of that paper sees `&amp;`. Scoped
   tightly: only `amp;` DIRECTLY after an ampersand, so "amplitude" and a lone
   `\&` are untouched, with a near-miss case in the fixture.
5. **Known cost of `&`-as-data.** A `.bib` field containing a real `tabular`
   would no longer align. Judged acceptable — no such witness in the 6000-source
   scan — but it is the honest price of regime A.
6. **A `copyright`-specific fix was implemented, then withdrawn.** Before the
   policy call, `copyright` was the one field in the cluster with a #73-style
   argument (no `.bst` emits it; no `FMT_SPEC` queries
   `ltx:bib-date[@role='copyright']`), and reading it `Verbatim` took 2605.01936
   7 → 6. Dropped once `&` became data everywhere — it would be a second
   overlapping mechanism in a file four agents are editing. *Reopen if:* a
   non-`&` special in `copyright` shows up in a sweep.
7. **`latexml_core/src/mouth.rs` is shared machinery.** The change is additive —
   one `bool`, one builder method, one arm in `catcode_of` — and inert unless a
   caller opts in; the only opt-in is the BibTeX path. Flagged because it is not
   my own binding.
8. **Renamed #405's `mouth::tokenize_percent_literal` → `tokenize_bib_literal`**
   (one call site). It now neutralizes `%` *and* `&`, so the old name would
   mislead. Trivial to revert if #405's author prefers the original name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,12 @@
     a percent that stays literal (BibTeX has no comments, and `%` is routine in
     an encoded URL). Measured over a 30,079-document arXiv sample: 62 more
     documents convert cleanly, 44 fewer carry errors.
+  - **An ampersand in a reference is an ampersand** — "Taylor & Francis" in a
+    `.bib` publisher, journal or booktitle used to report an error and print
+    "Taylor Francis". A `.bib` field's content is data, so the character is kept.
+  - **A reference exported through HTML no longer shows `&amp;` for `&`** — a
+    doubly escaped ampersand (`\&amp;`) in a `.bib` title, journal or booktitle
+    is decoded back to the one character the author wrote.
   - **amsrefs `\bib` values digest as live TeX** — `\MR{…}` came out as literal
     characters and `pages` rendered empty.
   - **MathReview / ZentralBlatt links are synthesized** — `mrnumber`/`zblno`
--- a/docs/SYNC_STATUS.md
+++ b/docs/SYNC_STATUS.md
@@ -64,6 +64,32 @@ session of their own. Re-verify a row before planning on it (rule 1).
 
 ## Current status
 
+- **2026-07-26 (later still) — a bare `&` in a `.bib` field is data (OXIDIZED_DESIGN #75).**
+  Seven 2605 witnesses carried `Error:unexpected:&` from `publisher` / `journal`
+  / `booktitle` / `author` / `copyright` ("Taylor & Francis"). Not a Rust-only
+  defect: same-host `latexmlc` raised the identical per-`&` count on all six
+  re-measured witnesses, and bibtex 0.99d + pdflatex agree (the `&` reaches the
+  `.bbl` under `plain` and `abbrvnat`; pdflatex stops with "Misplaced alignment
+  tab character &" and prints "Taylor Francis"). **A `.bib` field's content is
+  DATA** — authorized surpass-Perl and surpass-pdflatex, since LaTeXML reads
+  `.bib` directly and decides what reaches the tokenizer. Neutralized at the
+  same two seams #74 uses for `%`, and BOTH are required — the per-entry Mouth
+  (`Mouth::with_align_as_other`) and `mouth::tokenize_bib_literal` for handlers
+  that re-tokenize a stored raw field (`\bib@@title` recasing, name/date/pages
+  assembly, MR/Zbl). Measured: **2605.06249 3→0, 2605.03054 1→0, 2605.00462 1→0,
+  2605.08753 1→0, 2605.10409 1→0, 2605.01936 13→6, 2605.06624 4→3** (the
+  residuals are unrelated `undefined:` errors). `#` was checked and NOT included
+  — one bare `#` across all seven, in a JabRef `file` path already covered as an
+  unknown field. **Stacked on PR #405**: merge after it.
+  Also fixed, a different bug the neutralization does *not* reach: the doubly
+  escaped `\&amp;` / `{\&}amp;` / `&amp;`, an HTML entity that survived into
+  the `.bib` and printed as "&amp;" in Perl and pdflatex alike
+  (`undouble_escaped_ampersand`). Guards
+  `bib_bare_ampersand_is_literal_data`, `bib_bare_ampersand_leaves_live_markup_alone`
+  (the `\emph` / inline-math / space-form-accent boundary) and
+  `bib_escaped_amp_entity_decodes_to_one_ampersand`. Detail in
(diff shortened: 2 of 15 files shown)


dginev:
Superseded by #403, which absorbed this branch's commits, fixtures and guards **verbatim** and landed as `966b3b2de6`.

Nothing here is discarded. The joint two-treatment design keeps this PR's mechanism in its correct place — treatment 1 (reading `.bib`, be bibtex) neutralizes `% & #` at the mouth without altering the text; treatment 2 (synthesizing the `.bbl`, be pdflatex pass 2) escapes at the data→TeX boundary and covers `_` as well, since only the escaper can be math-aware.

The consolidated result is better than this PR reached alone: **193 → 0** across all 16 witnesses of the `_`/`%`/`&` clusters, with 2605.01936 going 13→0 (this branch reached 6) and 2605.06624 4→0 (measured 3 here). Recorded as OXIDIZED_DESIGN #74.

Issue #81, opened 2022-04-10
Note on improving "compiled replacements" via "cargo expand"


dginev:
I think it is instructive to describe a tiny piece of code I am upgrading today, as I am (again) increasing the coverage of the replacement compiler to get closer to latexml's.

Rust allows for some pleasant debugging of the code generation machinery, via `cargo expand`. To best benefit, we can work on isolated macro invocations. For example, I can create a dedicated `test.rs` file, containing the example I am improving, say:
```rust
use crate::package::*;
LoadDefinitions!(state, {
DefConstructor!("\\@@POSTSUPERSCRIPT InScriptStyle",r#"""
  <ltx:XMApp role="POSTSUPERSCRIPT" scriptpos="?#scriptpos(#scriptpos)(#scriptlevel)">
    <ltx:XMArg rule="Superscript">#1</ltx:XMArg>
  </ltx:XMApp>
  """#);
});
```
and store it under `rtx/rtx_package/src/package/pool/test.rs`, also adding a `mod test;` in `.../pool/mod.rs`. 

Then, from the terminal, we can ask to see this macro expanded via:
```
$ cargo expand package::pool::test
```

which currently prints out the following incorrect Rust code:

<details>

```rust
mod test {
    use crate::package::*;
    pub fn load_definitions(outer_stomach: &mut Stomach, state: &mut State) -> Result<()> {
        {
            {
                {
                    let (cs, params) = {
                        let st: &mut State = {
                            struct _Bound;
                            state
                        };
                        parse_prototype("\\@@POSTSUPERSCRIPT InScriptStyle", st)?
                    };
                    let compiled_replacement;
                    {
                        use rtx_core::definition::ReplacementClosure;
                        #[replacement = r#"""
  <ltx:XMApp role="POSTSUPERSCRIPT" scriptpos="?#scriptpos(#scriptpos)(#scriptlevel)">
    <ltx:XMArg rule="Superscript">#1</ltx:XMArg>
  </ltx:XMApp>
  """#]
                        struct _Dummy;
                        let tmp: Option<ReplacementClosure> = Some(Arc::new(
                            |document: &mut Document,
                             args: &Vec<Option<Digested>>,
                             props: &HashMap<String, Stored>,
                             state: &mut State| {
                                let mut savenode: Option<Node> = None;
                                document.absorb_string("\"Superscript\">", props, state)?;
                                if let Some(ref stored_digested) = args[0usize] {
[...]

PR #428, opened 2026-07-27
fix(engine): \meaning of a \chardef token reports the character, not "Register"

dginev:
Clears the largest unclassified fatal cluster in sandbox-arxiv-2605+2606: 17 papers whose first error is `unexpected:fi`. One cause, all 17.

## Diagnostic

Rust's `\meaning` had **no chardef arm**. A `CharDef` and a plain `Register` are both `Stored::Register` here, discriminated by `register_type`, so a chardef fell through a catch-all and rendered as the literal string `Register`. Perl dispatches on the concrete class — `ref` yields `...::Definition::CharDef`, which misses the `/register$/` arm — and builds `\char"<value>` at `Engine/TeX_Debugging.pool.ltxml` L166-168.

The dropped `"` is load-bearing. `bxcoloremoji.sty` (shipped by all 17 papers) recovers the character value with a delimited argument, `\def\bxce@do#1"#2\relax`. With no `"` in the string the argument scan runs away and swallows the `\fi\fi` of the enclosing `\AtEndOfPackage{…\@whilenum…}` loop (L1366-1386). Those `\fi` tokens then execute against an empty if-stack **from a macro body** — which is exactly why the locator reads `at Anonymous String` rather than a file:line, with a 1000-error cascade behind it.

## Approach

Port Perl exactly: test the chardef case first, emit `\char` + `"` + the decimal value.

Perl's two deviations from `tex.web` L22897-22899 — decimal instead of hex, and `\char` for `\mathchardef` — are inherited **deliberately**, and recorded as `KNOWN_PERL_ERRORS` #65. Perl's ternary keys off `$$definition{mathglyph}`, but `Core/Definition/CharDef.pm` L32-35 never sets that key, so the `\mathchar` arm is unreachable upstream. Our `Register` *does* carry a decoded `mathglyph`, so branching on it would have been a silent divergence; the comment at the fix site says so.

## Validation

Release build with dumps, `--preload=ar5iv.sty --path=…/ar5iv-bindings/bindings`, main file taken from each paper's own cortex `Processing content` line. Perl 0.8.8 (rev `0d02309d`) same host, verbose, ANSI-stripped.

| witness | Rust before | Rust after | Perl |
|---|---|---|---|
| 2605.03971 | 1002 | 2 | 50 |
| 2605.04451 | 1002 | 1 | 2 |
| 2605.09005 | 1002 | 1 | 51 |
| 2605.15128 | 1002 | 1 | 37 |
| 2605.16720 | 1002 | 1 | 2 |
| 2605.29156 | 1002 | 1 | 3 |
| 2605.29341 | 1002 | 8 | 11 |
| 2606.06712 | 1002 | 3 | 11 |
| 2606.07410 | 1002 | 1 | 12 |
| 2606.11290 | 1002 | 10 | 102 |
| 2606.11722 | 1002 | 1 | 29 |
| 2606.13769 | 1002 | 4 | 5 |
| 2606.14502 | 1002 | 1 | 2 |
| 2606.15753 | 1002 | 4 | 4 |
| 2606.18180 | 1002 | 1 | 1 |
| 2606.24256 | 1002 | 1 | 5 |
| 2606.26947 | 1002 | 1 | 2 |

All 17 went `rc=1` + `Fatal:TooManyErrors` to `rc=0`, with **zero** `unexpected:fi`. Perl never raised `unexpected:fi` on any of them, so the verdict is GENUINE-RUST-ONLY throughout; post-fix Rust is at or below Perl on every witness.

- Guard `meaning_chardef` (`latexml_oxide/tests/expansion/meaning_chardef.{tex,xml}`), hand-authored golden, independently confirmed against Perl. Red before the fix (`T1[Register]` — the runaway scan), green after (`T1[65]`).
- Full suite **1751 passed / 0 failed**.
- `cargo +nightly clippy --workspace --all-targets -- -D warnings` clean; `cargo +nightly fmt --all` applied.
- Lateral regression sweep over 62 unrelated papers: **62/62 identical** before vs after. Includes the two witnesses already named in this file's comments, `2110.11931` and `1908.01908` (0 errors both sides).

Note: PR #426 fixed none of these — all 17 still reproduced at `fc56b4d081`, which *is* #426.

### DIFF
--- a/docs/SYNC_STATUS.md
+++ b/docs/SYNC_STATUS.md
@@ -64,7 +64,52 @@ session of their own. Re-verify a row before planning on it (rule 1).
 
 ## Current status
 
-- **2026-07-27 (latest) — spconf.sty's `keywords` and `\twoauthors` were
+- **2026-07-27 (latest) — the `unexpected:fi` fatal cluster: `\meaning` of a
+  `\chardef` token returned the internal class name.** GENUINE-RUST-ONLY,
+  **18 papers, one cause.** Largest unclassified first-error cluster in the 186
+  `Fatal:TooManyErrors` papers of sandbox-arxiv-2605+2606:
+  `2605.{03971,04451,09005,15128,16720,29156,29341}`
+  `2606.{06712,07410,11290,11722,13769,14502,15753,18180,24256,26947}` — all 17
+  ship `bxcoloremoji.sty`. Rust `\meaning` had **no chardef arm**: CharDef and
+  plain Register are both `Stored::Register` (discriminated by `register_type`),
+  so a chardef fell through a catch-all and rendered as the literal string
+  `Register`. The dropped `"` is load-bearing — `bxcoloremoji.sty` L1373 recovers
+  the value with the delimited `\def\bxce@do#1"#2\relax`, so with no `"` the
+  argument scan runs away and swallows the `\fi\fi` of the enclosing
+  `\AtEndOfPackage{…\@whilenum…}` loop (L1366-1386). Those `\fi`s then executed
+  against an empty if-stack **from a macro body — hence the `at Anonymous String`
+  locator**, which is the tell that separates this from a source-level `\fi`.
+  Fixed faithfully per `TeX_Debugging.pool.ltxml` L166-168 (`\char` + `"` +
+  decimal). Measured, release + dumps, `--preload=ar5iv.sty`: **1002 -> 1-10
+  errors on all 17, zero `fi`, no fatal**; same-host Perl was 1-102 with no `fi`,
+  so post-fix Rust is at or below Perl on every witness. (Both caps must be named
+  or the deltas mislead: our 1002 is the tikz-raised 1000 cap, and Perl's lone
+  102 — 2606.11290 — is Perl's own `MAX_ERRORS`=100, so that one Perl total is
+  >=100 and unknown.) PR #426 fixed none of them (all 17 still reproduced at
+  `fc56b4d081`, which *is* #426). Guard `meaning_chardef`
+  (`latexml_oxide/tests/expansion/`). Perl's own two deviations from `tex.web`
+  L22897-22899 here (decimal not hex; `\char` for `\mathchardef`) are
+  deliberately inherited — recorded as `KNOWN_PERL_ERRORS` #65, which also warns
+  why Rust's populated `mathglyph` must NOT be used to revive the `\mathchar`
+  arm.
+  **Method note — first-error bucketing UNDER-counted this cluster.** Exactly 18
+  papers in 2605+2606 (of 60,513) ship `bxcoloremoji.sty`; 17 surfaced as
+  `unexpected:fi`, and the 18th, **2605.14271**, bucketed elsewhere because its
+  *first* error was `undefined:\SetTitleBoxVerticalShift` — yet it carried the
+  same two `fi` errors, was a real `Status:conversion:3`, and went **1002 -> 12**
+  (Perl 42) on the same fix. So a first-error histogram is a lower bound on a
+  cause's reach, not a census: confirm membership by the mechanism (here, "ships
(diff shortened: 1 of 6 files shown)


dginev:
Follow-up measurement: the cluster is **18 papers, not 17**.

Exactly 18 papers in sandbox-arxiv-2605+2606 (of 60,513) ship `bxcoloremoji.sty`. 17 surfaced as `unexpected:fi`; the 18th, **2605.14271**, bucketed as a singleton because its *first* error was `undefined:\hypersetup` — but it carried the same two `fi` errors and goes **103 → 1 error, fi 2 → 0** on this fix. Perl same-host: 30.

Worth stating explicitly since it cuts against the usual failure mode: a first-error histogram is a **lower bound** on a cause's reach, not a census. Membership was confirmed by the mechanism (ships the package), then re-measured.

dginev:
One caveat on the Perl column, so the numbers are not over-read:

**2606.11290's Perl figure (102) is capped.** Perl hit its own `Fatal:too_many_errors:100` (`Core/State.pm` default `MAX_ERRORS` = 100), so its true uncapped total there is ≥100 and unknown. The other 16 Perl runs completed without hitting the cap, so those are real totals.

The Rust *before* column (1002) is likewise a cap — ours is raised to 1000 by `tikz.sty`. Comparing 1002 vs 102 as if both were totals would be a cap artifact, not a divergence. The load-bearing comparison is the uncapped *after* column (1–10) against the 16 uncapped Perl totals, plus the `unexpected:fi` count going 2 → 0 on every witness — Perl never raised it at all.

dginev:
**Correction to my 18th-paper comment above** — I had used a heuristic main-file pick and it was wrong.

My detector chose the largest `.tex` containing `\documentclass`, which for 2605.14271 is `macro.tex` — an *include*, not the main file. Re-measured with cortex's own `Processing content` line (`main.tex`):

| 2605.14271 | errors | `fi` |
|---|---|---|
| Rust before | **1002** | 2 |
| Rust after | **12** | 0 |
| Perl | 42 | 0 |

Recorded cortex status was `Status:conversion:3` — a real fatal, not a marginal case. So the corrected figure is **1002 → 12**, not the 103 → 1 I posted; the paper is a full member of the cluster, making it **18 papers**.

Flagging the process point because it is the same failure mode the table in the PR body avoids: a wrong main-file pick manufactures plausible numbers. The 17 in the body were all picked from cortex's line, so they are unaffected.

dginev:
Rebased onto `main` @ `8a780703af` (post-#427) — now `72f8dbdd86`, single commit, `main` verified as ancestor.

**Conflict** was `docs/SYNC_STATUS.md` only, and purely additive: #427 and I each prepended a bullet to `## Current status`. Both entries kept in full. The only edit to #427's text was its same-day ordering marker, `(latest)` → `(later still)`, since my entry now sits above it — flagging that explicitly rather than changing someone else's words silently. No `CHANGELOG.md` conflict (I never touched it).

**Numbering verified, not assumed** — no collision. `KNOWN_PERL_ERRORS` #65 is mine (unique, follows #64, `uniq -d` finds no duplicate); `OXIDIZED_DESIGN` #82 is #427's, different file. I took no divergence number at all, correctly — this is a faithful port, not a divergence.

While verifying I did find a real instance of the trap you named: `OXIDIZED_DESIGN_DIVERGENCES.md` restated "Next free number" in **two** places, and the in-body copy had drifted stale at **#81** while the authoritative header had moved to **#83**. That is precisely how a number gets taken out from under someone. Fixed by making the in-body copy point at the header instead of restating a number, so there is one counter, not two. One line, and #427 is already merged so there was no branch to conflict with.

Suite on the new base: **1754 / 0**, exactly as you predicted. Both my `meaning_chardef` and #427's three `frontmatter_spconf_*` guards confirmed live and green. `fmt --check` and `clippy -D warnings` clean.

PR #430, opened 2026-07-29
feat(acmart): \Description becomes the figure's ARIA text alternative

dginev:
**Diagnostic.** acmart documents `\Description` as *"used instead of the image"* — a text alternative. It reached no output at all: the binding (a faithful port of Perl's `acmart.cls.ltxml` L78-86) emits `#1`, the **optional short** description, and digests then discards `#2`, the mandatory long one; `\Description{L}` produced nothing whatsoever. Upstream's own `TODO` there asks what to do with the short description. Recorded as `KNOWN_PERL_ERRORS` #66.

**Approach.** The two arguments map to the two ARIA slots a text alternative uses — `[short]` → `aria-label`, `{long}` → `aria-describedby` block; a lone description labels directly unless it carries markup an attribute cannot hold. Choosing between those slots is why the argument is read `Undigested`: the tokens are inspected for control sequences *before* anything expands. That also stops manufacturing errors from content `acmart.cls` L895 gobbles, which pdflatex never expands either. Divergence recorded as `OXIDIZED_DESIGN_DIVERGENCES` #83.

**Validation.** `110_acmart_description_aria` asserts at the **HTML** level that all three mapping rows hold, that markup survives, that every `aria-describedby` reference resolves, and that no footnote scaffolding leaks into the announced text — verified red on the pre-fix code. Full suite 1756/0; clippy and fmt clean. arXiv:2607.21760 went from `6 warnings; 1 error` with **zero** figure descriptions to `6 warnings` with **four**. `t/complex/acm_aria.xml` re-blessed — it previously matched Perl byte-for-byte and so certified the defect.

### DIFF
--- a/latexml_oxide/tests/complex/acm_aria.xml
+++ b/latexml_oxide/tests/complex/acm_aria.xml
@@ -6,7 +6,7 @@
   <resource src="LaTeXML.css" type="text/css"/>
   <resource src="ltx-article.css" type="text/css"/>
   <resource src="ltx-amsart.css" type="text/css"/>
-  <figure aria:labelledby="acmlabel1" inlist="lof" labels="LABEL:fig:one" xml:id="S0.F1">
+  <figure aria:describedby="acmlabel1" aria:label="Fly 1 and Fly 2 look identical" inlist="lof" labels="LABEL:fig:one" xml:id="S0.F1">
     <tags>
       <tag>Figure 1</tag>
       <tag role="autoref">Figure 1</tag>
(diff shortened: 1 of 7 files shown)


latexml_oxide/tests/complex/acm_aria.xml
```diff
@@ -6,7 +6,7 @@
   <resource src="LaTeXML.css" type="text/css"/>
   <resource src="ltx-article.css" type="text/css"/>
   <resource src="ltx-amsart.css" type="text/css"/>
-  <figure aria:labelledby="acmlabel1" inlist="lof" labels="LABEL:fig:one" xml:id="S0.F1">
+  <figure aria:describedby="acmlabel1" aria:label="Fly 1 and Fly 2 look identical" inlist="lof" labels="LABEL:fig:one" xml:id="S0.F1">
```
xworld21, line 9: I wasn't quick enough to properly check this, but the ARIA label on the figure tag itself will probably hide the caption from screen readers. Best suggestion that comes to mind is that label + description should be attached to the first image in the figure (and making sure that the `` tag eventually gets `@alt` and not `@aria-label`!). If the images already have an alt attribute, well... one has to wonder what the author was trying to do.

The acmart documentation is also a little puzzling, as they suggest putting active TeX content in the `alt` option, but as far as I know PDFs only allow pure text in there.

dginev: Their intention is for `\Description` to provide an alternative text for the image asset itself, but the way it gets deposited associated it with the figure. The code could improve to properly detect the image asset and set an attribute on it, but it's finnicky.

Their latest docs mention they've now migrated to using the `alt` key for `\includegraphics` , reinforcing that reading.

dginev: Also, being "quick enough to review" would be rather superhuman at the moment, the agentic loop is faster than the speed of human understanding.

The way I see it, refinements will happen closer to "the end" of the initial development mad dash, with most of the AI turned off.

PR #441, opened 2026-07-29
perf(core): one-borrow token read, duty-cycled cycle guard, pinned hot keys

dginev:
## What

A profiling-driven pass on the token-read hot path. On huge-token-stream documents (pgfplots witness `2405.14114`: 96% of its 23.9 s wall in digest, 200M+ tokens), the engine's own per-token safety guards were the #1 and #2 self-time functions — ~10% of the conversion — with another ~5% going to re-interning the same strings per call. Three output-neutral changes:

1. **One-borrow combined read** (`gullet.rs::read_internal_token_checked`). The per-token trio `read_resource_checkpoint` → `read_internal_token` → `cycle_guard_checkpoint` (up to three thread-local `RefCell` borrows per token) is now a single `GULLET` borrow shared by all four reader loops (`read_token`, `read_x_token`, `read_balanced`, `read_next_conditional`). Limit breaches Fatal via `#[cold]`-outlined helpers with messages, debug dumps, and "a breach consumes no token" semantics preserved verbatim. `read_balanced` keeps its comments-in-result behavior via a `CommentSink` parameter; the COMMENT/MARKER routing is shared (one closure), not duplicated per source.

2. **Duty-cycled cycle guard.** Once `progress` crosses the activation floor, the guard used to fingerprint every remaining token forever — pgfplots data plots legitimately run past even the raised 150M graphics floor. It now scans 2048 of every 16384 tokens, resetting the ring at each ON-window start. A genuine infinite loop is *persistent*, so detection stays guaranteed: worst case ~one duty period (~17k tokens) added latency, four orders of magnitude under the 400M token-limit backstop. The ON width (2048) covers ring capacity (`MAX_WINDOW×REPEAT = 1000`) plus scan cadence (`CHECK_EVERY = 256`) for every period, so a loop already running at window start is caught within that window. Verified live: `\def\y{}\def\x{\y\x}\x` still trips `Fatal:Timeout:Recursion` in 0.6 s. `raise_cycle_guard_activate` now also resets the ring, closing a raise-while-active seam where the phase formula shift could splice two ON-windows' histories.

3. **Pinned hot keys.** `Mouth` caches `source_sym` at construction, so `get_locator`/`get_locator_from_start` (per conditional, per box, every error path) build `Locator`s with zero interner probes (new `Locator::from_sym`). `Conditional::invoke` (`if_count`/`if_limit`/`tracingcommands`), `after_assignment` (`afterAssignment`), `assign_internal` (`global` prefix), and `Expandable::new` (`protected`/`outer`/`long`) use `pin!`-cached keys via new `_sym` siblings (`lookup_int_sym`, `remove_value_sym`, `get_prefix_sym`). The pre-existing `&str` variants now delegate to their `_sym` twins (`lookup_value`, `lookup_bool`, `lookup_int`, `remove_value`, `get_prefix`) — one body each, no duplication.

## Measured (same-session, interleaved best-of-3, release build)

| witness | before | after | Δ |
|---|---|---|---|
| 2405.14114 (pgfplots, digest-bound) | 23.85 s | 21.45 s | **−10.1%** |
| 1911.09517 (math-parse-bound) | 5.95 s | 5.88 s | −1% |
| 82-paper regression-corpus sweep (sum) | 192.2 s | 185.6 s | −3.4% (indicative) |

RSS flat. Post-patch profile: the three former guard functions (13.25% combined self-time) are one merged function at 9.01%, and interner `get_or_intern` dropped off the >1.2% list. The review-round cleanups (shared routing closure, delegating `&str` variants, `Expandable::new` pins) re-measured at parity-or-better against the mid-point binary (−3.4 to −4.1% interleaved).

## Output-neutrality + verification

- **Byte-identical output** vs the pre-patch binary on both witnesses (`diff -rq` clean), deterministic across runs.
- **Error-locator parity**: an error-producing document diffed old-vs-new — all TeX-source locators (`line N col M`) identical, error HTML byte-identical. (`arena::pin` deduplicates, so the cached `source_sym` is bit-identical to what per-call pinning produced.)
- Suite `cargo test --workspace --tests --no-fail-fast`: **1756 passed / 0 failed** (and re-run green on the final tree).
- `cargo clippy --workspace --all-targets`: clean. `cargo doc -p latexml_core`: 0 warnings. `cargo +nightly fmt --all` applied.

## Side effects, enumerated

- **Guard detection latency**: +≤16.4k tokens worst-case for the gullet cycle guard (analysis above); the stomach-side box guard, token/pushback limits, RSS fuse, and wall-clock watchdog are untouched.
- **Finite periodic bursts** (≥100 repeats that stop on their own inside an OFF window) may no longer Fatal — by design: they are not infinite loops.
- **`read_balanced` with a vanished runtime** (fatal-recovery corner): previously panicked on `.unwrap()`, now degrades as an exhausted opaque boundary (documented at the call site; unreachable in a healthy conversion).
- **New pub API** follows the established `_sym` naming convention; no existing signature changed.
- Internal Rust-source line annotations in error logs (`In latexml_core/src/state.rs:NNN`) shift by a line — cosmetic, inherent to any edit of that file.
- Dump/format boundary untouched: no changed path affects dump records (guards, locators, and prefix probes are runtime-only state).

## Profiled follow-ups (not in this PR)

- `is_noexpand_family` string probe — now the top discretionary residual (2.07%): record a "starts with `\special_relax`" bit per symbol at intern time (arena-side bitvec) → bit test in `meaning_key`.
- `install_definition` allocates + pins `"{cs}:locked"` per `\def`/`\let` (~0.7% + churn; Perl pays the same — needs a maintained side-set of locked syms).
- `if_stack`'s seven `&str`-keyed accesses in `conditional.rs` (would need four more `_sym` twins for ≤0.3%; the key name is also referenced by the dump filters).
- `assign_internal`'s Global-scope undo-frame walk (~4%) is faithful Perl State semantics; a typed `State` field for LaTeXML-internal counters needs dump-filter + `if_stack` review first.

Docs: `PERFORMANCE.md` audit-log 2026-07-29 (incl. box-specific profiling method notes) and `ARXIV_PERFORMANCE.md` hotspot #5.


dginev:
**Pre-merge review record.** Two independent adversarial reviews of the diff (attempting to refute behavior preservation) came back with one genuine finding, fixed in c51fb204ca: the duty-window phase-1 ring reset sat inside the token branch, so an Exhausted read (mouth boundary) landing exactly on an ON-window start skipped that window's reset — hoisted above the read. All other refutation attempts failed with evidence: ordering (progress/limits/comment-marker routing/pushback-before-mouth), borrow safety under the single held GULLET borrow, duty-cycle arithmetic incl. detection guarantees for all periods ≤ MAX_WINDOW, all four call-site mappings, Fatal-path string/dump parity, `Locator::from_sym` field mapping, and an exhaustive sweep for any `Mouth` construction that could desync `source`/`source_sym` (none exist; the field is private with no post-construction writes).

Verification on the final merged tree: suite 1760/0 locally + both CI legs, clippy/fmt/rustdoc clean, witness output byte-identical incl. error-locator parity, runaway loop still trips `Fatal:Timeout:Recursion` in 0.6 s, perf ratio re-confirmed interleaved (−5 to −6% vs the borrow-merge-only binary on 2405.14114). The earlier Linux CI failure was a stalled apt mirror during TeX Live install (zero tests ran); it passed on re-run.

PR #496, opened 2026-08-03
feat(diag): memory Fatals report what the document needs; disk staging loses the scary wording

dginev:
Two UAT-driven diagnostics improvements from the 131 MB witness runs.

## A memory Fatal now tells the user what the document needs

At \`--max-memory=26000\` the witness died with the bare \`Memory budget exceeded: RSS 19947 MB > cap 20447 MB\` — no hint that the cap is the 75% fuse of a flag the user can raise, nor that a ~23 GB core floor is a *known need* of a document this size, not an anomaly.

- The cooperative-fuse Fatal (\`Timeout:MemoryBudget\`) now names the fuse fraction and the \`--max-memory\` figure it derives from, and advises rerunning with a higher ceiling. The watchdog hard-kill line carries the same advice.
- **Only after such a Fatal** (clean runs stay quiet — reviewer directive), the end of the run adds \`Info:memory:peak\` with the kernel-tracked peak RSS (Linux \`VmHWM\`, macOS \`getrusage\`, Windows \`PeakWorkingSetSize\`): the honest lower bound on what THIS article needs. Emitted to stderr and recorded just above the log's \`Status:conversion:N\` tail (plain \`--log\` and archive-packed logs both).
- No "sufficient" figure is claimed: the fuse clamps the peak and the spill watermark itself derives from the ceiling, so the true requirement is only found by rerunning higher.

## "spilled" → "staged to disk"

\`459,579 segment(s) spilled\` reads as an accident, but writing segments/pages to disk when the document does not fit RAM is exactly what the streaming design intends. User-facing log/error text now says "staged to disk" / "disk-staged"; internal identifiers (\`spill_dir\`, \`render_spilled_page\`, \`spill_watermark_bytes\`, \`.latexml-spill-*\` temp dirs) keep their names. No test or tool asserted on the old wording.

Gates: 1880/1880 nextest, clippy \`-D warnings\`, rustdoc — all green.

### DIFF
--- a/latexml_core/src/document.rs
+++ b/latexml_core/src/document.rs
@@ -1703,10 +1703,10 @@ impl Document {
         emit_error(
           "spill",
           "unresolved",
-          &format!("a nested spilled segment could not be spliced ({elem})"),
+          &format!("a nested disk-staged segment could not be spliced back ({elem})"),
         );
         out.push_str(before);
-        out.push_str("<!-- latexml-oxide: LOST spilled segment -->\n");
+        out.push_str("<!-- latexml-oxide: LOST staged segment -->\n");
         rest = &from[elem_end..];
         continue;
       };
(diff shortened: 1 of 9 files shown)


dginev:
Third commit on this branch: the `[N]` math-progress markers now count conversion-wide. Streaming pass 2 builds a fresh `MathParser` per staged fragment, so the per-instance counter restarted at `[1]` in every fragment — useless for judging global position on a 500k-formula document. The ordinal is now a conversion-wide thread-local, reset once per conversion in `reset_conversion_notices` (called from `initialize_singletons`), so pooled `--server` threads do not inherit the previous document's count.

dginev:
Fourth commit: `Scan: DBStatus: N objects` now backs off exponentially — it logs only when the object count passes a power of two (~20 lines for a 254k-object document instead of one line per scanned page).

Issue #543, opened 2026-08-13
no way to assign non-strings from Rhai


xworld21:
While the new API admits typed LookupNumber, LookupString, there are no Assign* counterparts.

Also, there is no `LookupFloat/AssignFloat` pair (related to #542 if not already obvious).

xworld21:
PS: alternatively, the internal functions like `try_lookup_int` can also try parsing strings.

PPS: there is no internal `lookup_float`.

dginev:
Thank you, we should support all of these. Since the last few issues you've opened are direct interface gaps, I'll try to add some fixes soon.

PR #547, opened 2026-08-14
feat: PushValue list family + search-path bindings for Rhai (#540)

### LINKED ISSUE #540: PushValue not available in Rhai bindings
`PushValue` is not available. My use case is adding a directory to `SEARCHPATHS`, although I don't understand if `SEARCHPATHS` is even a thing here. I don't see code using `SEARCHPATHS` anywhere.

dginev:
**Diagnostic.** The runtime Rhai bindings had no `PushValue` (nor the rest of the Perl `@values` list-op family). Separately, the reporter's goal — appending a directory to `SEARCHPATHS` — cannot go through `PushValue`: the Rust port keeps input search paths in a dedicated `State.search_paths` field (read by `find_file`), not a value-table key, so `PushValue("SEARCHPATHS", dir)` would be a silent no-op.

**Approach.** Register `PushValue`/`PopValue`/`UnshiftValue`/`ShiftValue` (Perl `Package.pm` L265-279 -> `State.pm` L218; global auto-vivify, no scope arg) wrapping the existing `latexml_core::state` functions, plus dedicated `PrependSearchPath`/`AppendSearchPath` over `state::search_paths_push_front`/`add_search_path`. The list ops are **type-general**: `PushValue`/`UnshiftValue` take any Rhai value and marshal it to the matching `Stored` (string/int/float/bool/`Tokens`/digested handle), and `PopValue`/`ShiftValue` return it with its type preserved rather than stringified — Perl's `PushValue` is untyped and the State layer already holds any `Stored` (`Dimension`/`Glue`/`Font` have no Rhai representation, so they stay unreachable). `GRAPHICSPATHS` **is** value-table-backed, so `PushValue` reaches it. No new State machinery.

**Validation.** Guards `pushvalue_family_and_search_paths` (string family + search-path field + `GRAPHICSPATHS`; confirms the `SEARCHPATHS` value-table key stays absent) and `pushvalue_preserves_types` (int/float/bool/`Tokens`/digested round-trip with type preserved — `type_of(popped) == "i64"`). `API.md` doc-gate regenerated (93->99 functions). Full suite `cargo nextest run --workspace` 1931/1931, 0 skipped; `clippy --workspace --all-targets -D warnings` clean; fmt applied.

Closes #540

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@
     `LookupNumber`/`LookupBool`/`LookupString`; and a new `LookupFloat`/`AssignFloat`
     pair (plus internal `state::lookup_float`) carries fractional values a
     `Number` would truncate (#543).
+  - **Runtime Rhai bindings gain the list-value family and search-path control.**
+    `PushValue`/`PopValue`/`UnshiftValue`/`ShiftValue` expose the Perl `@values`
+    list-op set (e.g. for `GRAPHICSPATHS`), and `PrependSearchPath`/`AppendSearchPath`
+    add an input directory — the Rust port keeps `SEARCHPATHS` in a dedicated field,
+    not the value table, so these reach file resolution where `PushValue` would not (#540).
 
 ## [0.7.5] (Rhai binding API; bibliography content recovery; wider math coverage; large-document memory; default-CSS sync; ar5iv corpus fixes)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,11 @@
     `Number` would truncate (#543).
   - **Runtime Rhai bindings gain the list-value family and search-path control.**
     `PushValue`/`PopValue`/`UnshiftValue`/`ShiftValue` expose the Perl `@values`
-    list-op set (e.g. for `GRAPHICSPATHS`), and `PrependSearchPath`/`AppendSearchPath`
-    add an input directory — the Rust port keeps `SEARCHPATHS` in a dedicated field,
-    not the value table, so these reach file resolution where `PushValue` would not (#540).
+    list-op set (e.g. for `GRAPHICSPATHS`), carrying any Rhai-representable value —
+    string/int/float/bool/`Tokens` — with the type preserved across a push/pop
+    cycle. `PrependSearchPath`/`AppendSearchPath` add an input directory: the Rust
+    port keeps `SEARCHPATHS` in a dedicated field, not the value table, so these
+    reach file resolution where `PushValue` would not (#540).
 
 ## [0.7.5] (Rhai binding API; bibliography content recovery; wider math coverage; large-document memory; default-CSS sync; ar5iv corpus fixes)
 
--- a/latexml_contrib/src/script_bindings/API.md
+++ b/latexml_contrib/src/script_bindings/API.md
@@ -35,11 +35,12 @@ closure-registered functions; only the accepted types are.
 
 Called at the top level of a binding file, or from inside any body.
 
-93 functions, 124 calls.
+99 functions, 130 calls.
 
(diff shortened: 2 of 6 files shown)


xworld21:
Just a quick word: I managed to work around the SEARCHPATHS issue by leveraging the import.sty binding (although it happens to have the wrong semantics!).

How are you handling the typing issue? Does this just push/pop/shift/unshift strings (which may be fine for now)?

dginev:
Good question. Currently:

- **`PushValue` / `UnshiftValue` take a string** and store it as `Stored::String` — so the push side is strings-only, which (as you say) is fine for the path/option-list cases these usually serve.
- **`PopValue` / `ShiftValue` return a Rhai `Dynamic`**: `String → string`, `Bool → bool`, a digested handle passes through, and anything else comes back as its string form. So they don't *lose* type for strings/bools, but a non-scalar round-trips as text.

The underlying `state::push_value` is generic (`Into<Stored>`), so the State layer already holds typed list items — only the binding narrows to strings. If you get a use case for pushing non-strings, a typed push family mirroring the typed `Assign*` set from #543/#544 (`AssignNumber`/`AssignFloat`/`AssignBool`) is an easy follow-up — happy to add `PushNumber`/`PushFloat`/… if you'll use them. Strings-only for now otherwise.

On the workaround: yeah, `import.sty` is the wrong tool here — `\import`'s path handling is deliberately *scoped* (saved/restored around each subfile body via `\lx@save@paths`/`\lx@restore@paths`) and canonicalizes relative to `SOURCEDIRECTORY`, so a path added that way is local to the group and munged. `PrependSearchPath`/`AppendSearchPath` in this PR are the persistent, global way. If you think `import.sty` itself diverges from Perl's `import.sty.ltxml` (rather than just being the wrong tool for this), a separate issue with a minimal repro would be very welcome.

dginev:
(previous reply is by Opus, as should be obvious by the reply style...

and I see it is getting a bit lazy, I will prod it to generalize to dynamic typing a-la perl)

dginev:
Update: went ahead and generalized it in this PR rather than deferring. `PushValue`/`UnshiftValue` now take any Rhai value and store the matching `Stored` (string/int/float/bool/`Tokens`/digested handle); `PopValue`/`ShiftValue` return it with its type preserved (a popped int is an `i64`, not `"7"`). `Tokens` round-trips cleanly — e.g. `PushValue("k", TokenizeInternal("\\alpha"))` then `PopValue("k")` gives you back the tokens. The only unreachable `Stored` variants are the ones with no Rhai value at all (`Dimension`/`Glue`/`Font`/…). Guard: `pushvalue_preserves_types`.

dginev:
@xworld21 — thanks for the nudge on the general point; it was the right call. The list ops are no longer strings-only: `PushValue`/`UnshiftValue` take any Rhai value and `PopValue`/`ShiftValue` hand it back with its type intact, so `Tokens`, ints, floats, bools and digested handles all round-trip faithfully — à la Perl's untyped `PushValue`. The only things you can't push are the `Stored` variants with no Rhai representation at all (`Dimension`/`Glue`/`Font`/…), which a script has no way to construct anyway.

So both halves of #540 are covered — the `@values` family for value-table lists (e.g. `GRAPHICSPATHS`), and `PrependSearchPath`/`AppendSearchPath` for the input-directory case that `SEARCHPATHS`'s dedicated field needs. Merging once CI is green.

xworld21:
> yeah, `import.sty` is the wrong tool here — `\import`'s path handling is deliberately _scoped_

For reference, I reported the problem with import at https://github.com/brucemiller/LaTeXML/issues/2866 (and `\lx@append@path` does the job of `PushValue`, actually, and can take absolute paths if one fixes the current import bug).

PR #648, opened 2026-08-18
fix #638: don't leak LaTeXML font attributes onto raw xhtml markup

### LINKED ISSUE #638: font encoding gets attached to raw html attribute (?)
Baffling. Compile
```latex
\fontencoding{T1}\selectfont
\bmlRawHTML{<div>encoding leaked</div>}
```
with binding
```Rhai
DefConstructor("\\bmlRawHTML{}", |document, rawhtml| {
  let nodes = ParseXML("<div xmlns='http://www.w3.org/1999/xhtml'>" + ToString(rawhtml) + "</div>");

  if nodes.len > 0 {
    document.openElement("ltx:rawhtml");
    document.insertXML(nodes[0].children());
    document.closeElement("ltx:rawhtml");
  }
});
```
(The binding itself should change the font encoding here, in a group, but I am simplifying to make the issue more obvious.)

The output is
```xml
<rawhtml>
  <xhtml:div encoding="T1">
encoding leaked      </xhtml:div>
</rawhtml>
```

Clearly the `@encoding` attribute should not be there. In fact, I have never seen that attribute in perl, is there a reason for saving it into the file?

dginev:
**Diagnostic.** The font `encoding` is a FontMap-lookup key used *during* digestion to decode unicode — meaningless as an output attribute, and no `ltx:` element declares it. `Font::relative_to` still placed it in the relativized attribute set, where it normally vanished (`canHaveAttribute` false everywhere) but leaked onto raw xhtml a binding splices in via `insertXML`: the schema's `htmlElement { attribute * }` wildcard makes `canHaveAttribute('xhtml:div','encoding')` true, so `\fontencoding{T1}\selectfont` + a `\bmlRawHTML`-style binding produced `<xhtml:div encoding="T1">`.

**Approach.** Omit `encoding` from `Font::relative_to` (`common/font.rs`), so `@encoding` is never emitted anywhere. Supersedes the earlier namespace-gate approach per review (thread below) — the leak reproduces identically in a freshly-built upstream `brucemiller/LaTeXML` master, and no `ltx:` element ever accepts `@encoding`.

**Validation.** Guard `raw_html_insertxml_does_not_leak_font_encoding`; full suite 2052/0; clippy clean. (The only `@encoding` in output is MathML `<annotation encoding="…">`, set explicitly — unaffected.)

Closes #638

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,15 @@
     typed value, `()` when absent) and `document.getProperties()` (the whole map)
     read the same construction-time property map, matching Perl, which hands
     `$whatsit->getProperties` to a CODE replacement (`Constructor.pm:137`) (#635).
+  - **The font `encoding` is no longer emitted as an output attribute.** The
+    font-encoding property is a FontMap-lookup key used *during* digestion to decode
+    unicode; it is meaningless afterwards and no `ltx:` element declares it. It was
+    still placed in the relativized font-attribute set, where it normally vanished
+    (no element accepts it) but leaked onto raw xhtml a binding splices in via
+    `insertXML` — `\fontencoding{T1}\selectfont` + a `\bmlRawHTML`-style binding
+    produced `<xhtml:div encoding="T1">` (the div's `attribute *` schema wildcard
+    accepts anything). `Font::relative_to` now omits it, so `@encoding` appears
+    nowhere (#638).
   - **A shared author-email line distributes across the authors instead of bunching
     on the last one.** `\email{a@x, b@y, c@z}` (or a single `\email` covering several
     authors) previously attached every address to whichever creator was open when the
--- a/latexml_core/src/common/font.rs
+++ b/latexml_core/src/common/font.rs
@@ -1264,15 +1264,13 @@ impl Font {
         }),
       );
     }
-    if is_diff(self.encoding.as_ref(), other.encoding.as_ref()) {
-      result.insert(
-        "encoding".to_string(),
-        (self.encoding.as_ref().unwrap().to_string(), Font {
-          encoding: self.encoding.clone(),
-          ..Font::default()
-        }),
-      );
-    }
+    // #638: the font `encoding` is a FontMap-lookup key consumed DURING digestion
+    // (unicode decoding), not a presentation property — it is meaningless as an
+    // output attribute and no ltx element even declares it, so Perl emits no
+    // `@encoding` anywhere. Emitting it here only ever surfaced (spuriously) on a
+    // foreign `xhtml:*` element whose `attribute *` wildcard accepts anything.
+    // Omit it from the relativized attribute set entirely (Perl `Font.pm` keeps
+    // the analogous line, but the value never reaches real output there).
     if is_diff(self.language.as_ref(), other.language.as_ref()) {
       result.insert(
         "xml:lang".to_string(),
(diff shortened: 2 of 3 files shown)


xworld21:
I think the font encoding property of the font should simply be omitted, because it's not relevant after tokens have been digested. Also, you should not hardcode `is_xhtml_element`, as the issue would then pop up with any other schema change that accidentally allows the `@encoding` attribute. Perl does not generate an `@encoding` attribute anywhere as far as I know.

xworld21:
> Same-host Perl 0.8.8 emits the identical `<xhtml:div encoding="T1">`, so this is a **surpass** (SHARED-FAILURE)

I can't reproduce this issue with LaTeXML 0.8.8. Never seen `@encoding` before.

dginev:
@xworld21 You're right that `is_xhtml_element` was the wrong layer — I've reworked the fix to simply drop `encoding` from `Font::relative_to`'s attribute set (a FontMap-lookup key, not a presentation property), so `@encoding` is now never emitted anywhere and there's no namespace special-casing.

On the reproduction — I *do* get it from a **freshly-built current `master`** (`ee4b364`). The missing ingredient is a raw-HTML binding that `appendTree`s a foreign element while a non-default `\fontencoding` is active:

```perl
# myraw.sty.ltxml
package LaTeXML::Package::Pool;
use LaTeXML::Package; use LaTeXML::Common::XML;
DefConstructor('\myraw{}', sub {
  my ($document, $arg) = @_;
  my $xml = LaTeXML::Common::XML::Parser->new()->parseChunk(
    "<div xmlns='http://www.w3.org/1999/xhtml'>".ToString($arg)."</div>");
  $document->openElement('ltx:rawhtml');
  $document->appendTree($document->getElement, $xml) if $xml;
  $document->closeElement('ltx:rawhtml');
});
1;
```
```latex
\documentclass{article}\usepackage{myraw}
\begin{document}\fontencoding{T1}\selectfont\myraw{leaked}\end{document}
```
→
```xml
<rawhtml><xhtml:div encoding="T1">leaked</xhtml:div></rawhtml>
```

The chain:

1. `\fontencoding{T1}\selectfont` sets the font encoding to `T1`; the `ltx:rawhtml` insertion point carries it, and `openElementAt` → `setNodeFont` stamps that font onto the parsed `xhtml:div` (`_font`).
2. At finalization, `Font::relativeTo` (`Common/Font.pm`, the `isDiff($enc,$oenc) ? (encoding => …)` clause) puts `encoding => 'T1'` in the pending declaration, and `finalize_rec` (`Core/Document.pm` L385–402) applies each pending attribute wherever `canHaveAttribute($qname,$attr)` is true.
3. `canHaveAttribute('xhtml:div','encoding')` returns **1** — `htmlElement = element xhtml:* { attribute * … }` (`LaTeXML-misc.rnc`) registers `tagprop{xhtml:*}{attributes} = {*, *:*}` (anyName → `filterNames('*','*:*')`, `Model/RelaxNG.pm`). Whereas `canHaveAttribute('ltx:text','encoding') = 0`.

That asymmetry is why it's invisible in normal documents: no `ltx:` element declares `@encoding`, so the encoding diff is silently dropped everywhere — it only ever lands on a foreign `xhtml:*` element (whose `attribute *` wildcard accepts anything), and only when raw markup is spliced in under a differing `\fontencoding`.

Probe on the same build, inside `\myraw`:
```
canHaveAttribute(xhtml:div,encoding)=1   (xhtml:div,font)=1   (ltx:text,encoding)=0
tagprop{xhtml:*}{attributes} = [*, *:*]
rawhtml insertion-point font encoding = T1
```

Since `@encoding` isn't a meaningful output attribute anywhere, our fix just omits it from the relativized set. Happy to file the same upstream if you'd like `relativeTo` to drop it in Perl too.

dginev:
Quick follow-up: to make sure omitting `encoding` from `relative_to` doesn't blur font *identity*, I added a test that pins the three-way distinction for two fonts differing only in encoding:

- **unequal** and distinctly hashable (so `_font` / `node_fonts` still keep a T1 span and an OT1 span apart);
- **`distance` = 0** — encoding never opens a font wrapper (faithful to Perl's `distance`, which skips it);
- **`relative_to` = empty** — no output attribute.

So the two are still tracked as different fonts; they just no longer differ by a spurious `@encoding`. (`encoding_is_font_identity_but_not_distance_or_output` in `common/font.rs`.)

xworld21:
> Happy to file the same upstream if you'd like `relativeTo` to drop it in Perl too.

Thanks, yes. If this is happening upstream, it's also a bug there (apart from the digestion issue, the `@encoding` attribute appearing even without ever being mentioned in the schema is very suspicious, it should either have a defined semantics, or not exist at all).

dginev:
Btw I should stop calling the perl version "upstream" since it isn't quite that anymore. "perl latexml" seems clearest.

